PR #296, opened 2016-02-16
django 1.9.2 + requirements update closes #293

### LINKED ISSUE #293: Django 1.9
Current stable version of Django is 1.9.2.
Django 1.9 has some nice features (new lookups, new core features, improved template engine etc) and bug fixes.
Do you have any plans to support Django 1.9?
Also, why are you using `django-international`? This package is pre-alpha and last update was in 2014. This can create some problems for update to newer version of Django.

Any reasons why this package depends of Django 1.8.3? The latest version of Django 1.8 is 1.8.9.

krodyrobi:
@PressLabs/presslabs  CR

Since I've went through to updating the dependencies I've also looked over what would break in Django 1.10 and the only thing that is left from this PR is optparse / argparse from the Command parsing 1 occurrence, and nothing else that pops to eye.

- [ ] stage
- [ ] pypi
- [ ] git tag
- [ ] live

### DIFF
--- a/requirements/common.txt
+++ b/requirements/common.txt
@@ -1,16 +1,16 @@
-django==1.8.9
-djangorestframework==3.1.3
+django==1.9.2
+djangorestframework==3.3.2
 django-international==0.0.6
-django-fsm==2.2.1
-django-filter==0.10.0
+django-fsm==2.3.0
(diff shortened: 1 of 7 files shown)


zalmanu:
LGTM

PR #307, opened 2016-02-25
Total decimal points fix

bogdanpetrea:


silver/tests/unit/test_invoice.py
```diff
@@ -88,3 +90,36 @@ def test_cancel_issued_invoice_with_related_proforma(self):
...
+
+        assert self._get_decimal_places(invoice.total) == 2
+
+    def test_invoice_total_before_tax_decimal_places(self):
+        invoice_entries = DocumentEntryFactory.create_batch(3)
+        invoice = InvoiceFactory.create(invoice_entries=invoice_entries)
+
+        invoice.sales_tax_percent = Decimal('20.00')
+
+        assert self._get_decimal_places(invoice.total_before_tax) == 2
```
zalmanu: The value of the total should be tested as well, not just the no. of decimals.

zalmanu:
I see that there are two issues solved in this PR. My concerns are: 

1) Are all the changes through which a proforma/invoices goes logged? Some tests for this issue would be great.
2) We should also test the template generation. We should make sure that errors are not created after generating the complete template used for creating the `pdf` file.

silver/tests/unit/test_invoice.py
bogdanpetrea: :+1: There are already some tests that cover the total in test_generate_docs.py and spec tests for Invoices and Proformas.

bogdanpetrea:
@zalmanu: I guess I could visually check for errors. Not sure what could go wrong here `{{ entry.total|floatformat:2 }}`.

silver/tests/unit/test_invoice.py
rciorba: @bogdanpetrea since those tests failed to catch this bug till now, I'm with @zalmanu on this one.

bogdanpetrea: @rciorba: Those tests failed to catch this bug because they were not checking the number of decimals. They were only checking if the total amount was calculated correctly, which it was and still is. But I wrote tests for the number of decimals in this PR.

vtemian:
LGTM

PR #308, opened 2016-02-29
No numbers for draft documents (closes #285)

### LINKED ISSUE #285: Do not assign numbers to drafts
`Proformas` and `Invoices` should not get numbers when they are saved in `draft` state. Instead they should get number and series when they are issued (of course, if they don't have a manually assigned ones).

bogdanpetrea:


krodyrobi:
LGTM.

But to keep it consistent we should manually wipe no for active drafts.

PR #311, opened 2016-03-22
Removed pinned versions from requirements.txt (fixes #310)

### LINKED ISSUE #310: Lax requirered versions
This is an installable app. We should not pin versions in `requirements.txt` unless there is really a specific reason to do this. Some people may be in a situation in that they cannot install `silver` because they cannot move to the specific of a required package.

calind:
The only required package version is `django-jsonfield==0.9.15`

### DIFF
--- a/requirements/common.txt
+++ b/requirements/common.txt
@@ -2,20 +2,20 @@
 # dependency==x.x  # we really need this
 # DON'T do it the file will be parsed on deploy
 
-django==1.8.9
-djangorestframework==3.3.2
-django-international==0.0.6
-django-fsm==2.3.0
-django-filter==0.12.0
-djangorestframework-bulk==0.2.1
-django-livefield==2.4.0
+django>=1.8
+djangorestframework>=3.3.0
+django-international
+django-fsm>=2.3.0
+django-filter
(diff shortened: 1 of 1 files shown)


vtemian:
I really think that this is a bad idea, because when a maintainer from those packages will break the compatibility (eg: `djangorestframework`) will be frustrating to develop  or try something on silver.

krodyrobi:
I don't agree with removing the versions as some might break backwards comparability anytime. Better explicit than implicit imho.

rciorba:
As the others said, I think this is a bad ideea. I actually recall discussing this less than a month ago and deciding we want to pin EVERYTHING, because if we don't deployments won't be repeatable.

calind:
This is an installable package, it is not a deployment unit and I think this is the first thing to consider.

What actually happend is that this has django pinned to 1.8.9 and the main app had it pinned to 1.8.11 (which is the latest security release). After pip resolving the dependencies, it settled for django 1.8.9 leaving the main app in dubious security state.

As a general idea is think it's bad practice to pin versions here. The proper place to do it is in the main app.

rciorba:
@calind OK then, LGTM

PR #316, opened 2016-05-19
Cache billing documents name

bogdanpetrea:


silver/utils/decorators.py
```diff
@@ -0,0 +1,75 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from django.core.cache import cache
+
+
+def cached_method(unique_attribute=None):
```
vtemian: Instead of `unique_attribute`, `cache_key` is more explicit.

silver/utils/decorators.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+    A decorator for caching method results in the Django context
```
vtemian: What django context? views? models? migrations?

silver/utils/decorators.py
```diff
@@ -0,0 +1,75 @@
...
+    :param unique_attribute: Name of attribute used in the cache key
+    """
+    def decorator(func):
+        def func_wrapper(self, *args, **kwargs):
+            """
+            :param invalidate_cache: If True the cache is being invalidated
```
vtemian: `invalidate` should be enough because you are in a cache context. Repeating cache is redundant.

silver/utils/decorators.py
```diff
@@ -0,0 +1,75 @@
...
+            :param set_cache: If True the cache will be set if there is no cache
+                              If invalidate is True, the cache will be reset
```
vtemian: Redundant explication.

silver/utils/decorators.py
```diff
@@ -0,0 +1,75 @@
...
+            :param bypass_cache: If True the original method will be called
```
vtemian: `bypass` is enough

silver/utils/decorators.py
```diff
@@ -0,0 +1,75 @@
...
+                                 It can interact with the other parameters
+            :param return: Value of method (either cached or not)
+
+            !!! Warning: the *args and **kwargs parameters are considered when
+            creating the cache key
+            """
+
+            invalidate = kwargs.pop('invalidate_cache', False)
```
vtemian: Try to put them in function's signature. It's cleaner this way.

silver/utils/decorators.py
```diff
@@ -0,0 +1,75 @@
...
+            set_cache = kwargs.pop('set_cache', True)
+            bypass = kwargs.pop('bypass_cache', False)
+
+            if not unique_attribute:
+                unique_key = id(self)
```
vtemian: Usually a hash function is called. The id of the object is it's identifier, during the lifetime of the object. Each class that inherit from `object` should have an `__hash__` method. I suggest to use that instead of `id`

silver/utils/decorators.py
```diff
@@ -0,0 +1,75 @@
...
+from django.core.cache import cache
+
+
+def cached_method(unique_attribute=None):
+    """
+    A decorator for caching method results in the Django context
+    :param unique_attribute: Name of attribute used in the cache key
+    """
+    def decorator(func):
+        def func_wrapper(self, *args, **kwargs):
```
vtemian: `self` may be misleading because is used to refer to the current class instance. Here you are not working in the interior of a class, but in a function. Maybe `obj` is better, of `model`.

silver/utils/decorators.py
```diff
@@ -0,0 +1,75 @@
...
+            key = "{}-{}-{}".format(
+                self.__class__.__name__, func.__name__, unique_key
+            )
+
+            cache_is_fresh = False
+
+            if invalidate:
+                if set_cache:
+                    # reset cache
+                    cache.set(key, func(self, *args, **kwargs), None)
```
vtemian: What will happen if the result is another object? Is the cache backend able to set/retrieve it?

vtemian:
This is really nice and interesting work. Good job :D. Maybe we can polish the API a little bit and extract in a separate django package.

A part that is confusing to me is the `invalidate`, `set`, `bypass` logic. I don't see the point to mix those parameters and get a new functionality.

I'm thinking more like `update` and `purge`. Anyway, if a method is decorated with this decorator, you'll want to be cached, right? Other operations, which you can do, will be only to purge it, or update it.

If `update` and `purge` are `False`, `cache.set_or_get` will be called. If `update` is `True`, then you delete the key and reset it. On purge, you just delete it.

I'm thinking more and more if `purge` is a useful functionality...I don't see the point to purge a result, just to update it...

silver/utils/decorators.py
bogdanpetrea: As far as I know, the cache backend doesn't really care what type or what the stored value is. It will do what you tell it.

silver/utils/decorators.py
bogdanpetrea: Good point!

silver/utils/decorators.py
bogdanpetrea: I remember there was a good reason why I did this. I think I couldn't properly pass parameters to the decorators or something.

silver/utils/decorators.py
bogdanpetrea: Fair enough. Maybe the usage of `id` could be a last resort?

silver/utils/decorators.py
bogdanpetrea: That's what I initially thought as well. But then I thought there's a better chance that the original method might use a paramater named `bypass` than a `bypass_cache` named one. And then there's the other parameters that end up in `_cache`.

silver/utils/decorators.py
bogdanpetrea: Same explanation as for the `bypass_cache` parameter below.

silver/utils/decorators.py
bogdanpetrea: Just wanted to point out that invalidate must be used as well to reset the cache. I guess you're right.

silver/utils/decorators.py
bogdanpetrea: That's what I initially wanted to use, but then I thought that's not the final key that is being used and unique_attribute suggests you should be using a unique attribute. But you might be right.

silver/utils/decorators.py
bogdanpetrea: I'll rephrase that.
"A decorator for caching method results, using the Django low level cache API"

bogdanpetrea:
Thanks! We could do that.

I chose this set of parameters because I thought it would offer more maneuverability for whomever needs it.
`set` is `True` by default, which means that calling the method with `invalidate=True` would be the same as `update=True`.
I feel that `update` somehow incudes the `purge` functionality and it wouldn't be appropriate to have them in the same method call (e.g. `update=True, purge=False`). That doesn't mean it's a bad thing though. I'll think about it.

silver/utils/decorators.py
calind: I think `invalidate_cache` is better suited, considering that we are adding it to the original function parameter list. When reading the function call `foo(invalidate_cache=True)` is more readable than `foo(invalidate=True)`.

Aside from this injecting parameters into function signature is not the most elegant way to do cache management. Maybe there is another way.

calind:
I have the feeling that this PR introduces more complexity than necessary. Caching is always hard to get right. Are you sure that we cannot optimize the database access patterns? There is a pretty nice write-up in Django's docs https://docs.djangoproject.com/en/1.8/topics/db/optimization/. Also there is https://cache-machine.readthedocs.io/en/latest/ which caches db objects.

PR #318, opened 2016-06-15
python manage.py generate_docs --date throws an error

oucsaw:
We've been looking at using silver for our charging, and one of the things we wanted to do was simulate 
generating documents over a few months. So we set up some subscriptions, starting on 2016-01-01, and then ran:

```
 python manage.py generate_docs --date=2016-01-01
```
However, instead of offering a caffeinated beverage, we received a trace back:
```
<snip>
  File "/srv/silver/local/lib/python2.7/site-packages/silver/management/commands/generate_docs.py", line 57, in handle
    '%Y-%m-%d').date()
TypeError: must be string, not datetime.datetime
```

This seems to be because when adding the  date command line option in the add_arguments class method, it is converted from a string to a datetime.datetime object by the date function within the silver/management/commands/generate_docs.py module, but the handle method expects  options['billing_date'] to be a string.

This pull request adds a series of tests to check that the generate_docs command parses its arguments correctly, and then fixes the above issue by assuming that options['billing_date'] is either None, a datetime.date object or a string and acting accordingly. (The assumption of the string value may only be required for the test functions in silver/tests/commands/test_generate_docs_args.py; however rather than rewrite all of these, and find out this is required elsewhere, I've gone for the safe/lazy option to leave these tests alone - I'm happy to change the tests and remove the check for an input of string should this be felt a better option).

### DIFF
--- a/silver/management/commands/generate_docs.py
+++ b/silver/management/commands/generate_docs.py
@@ -51,8 +51,6 @@ class Command(BaseCommand):
         translation.activate('en-us')
 
         billing_date = options['billing_date']
-        if type( options['billing_date'] ) == StringType:
-            billing_date = date(options['billing_date'])
 
         docs_generator = DocumentsGenerator()
         if options['subscription_id']:
--- a/silver/management/commands/generate_docs.py
+++ b/silver/management/commands/generate_docs.py
@@ -16,6 +16,7 @@
 import logging
 import argparse
 from datetime import datetime as dt
+from types import StringType
 
 from django.core.management.base import BaseCommand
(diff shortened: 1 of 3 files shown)


calind:
Hi @oucsaw and thanks for your contribution. The proper way to do it is indeed changing the tests since the `billing_date` is a string only in tests.

oucsaw:
Hi @calind, I've now removed the string check from generate docs and updated the tests to reflect this

bogdanpetrea:
Nice work @oucsaw, LGTM.

PR #320, opened 2016-06-24
Resolve minor issue creating subscriptions with descriptions

oucsaw:
When creating a subscription with a description, the description was not being saved (and as subscriptions can't be patched other than meta data there was no way of adding this field via the API).

This commit updates:
 * silver/tests/spec/test_subscription.py
   * adding two functions to check that subscriptions created with a reference and description respectively, record this information
 * silver/api/serializers.py
   * fix so JSONSerializerField.to_internal_value that it passes hooks/pre-commit (the script was complaining about 'silver/api/serializers.py:93:17: W503 line break before binary operator', and I've moved the 'and' onto the previous line to resolve this)
   * added 'description' to the list of files in the SubscriptionSerializer so that you can create subscriptions with descriptions

The test to check the reference when adding a subscription is probably
unnecessary (as it worked before), but seems to do no harm.

### DIFF
--- a/silver/api/serializers.py
+++ b/silver/api/serializers.py
@@ -89,8 +89,8 @@ class JSONSerializerField(serializers.Field):
         if not data:
             return data
 
-        if (data is not None and not isinstance(data, dict)
-                and not isinstance(data, list)):
+        if (data is not None and not isinstance(data, dict) and
+                not isinstance(data, list)):
                     raise ValidationError("Invalid JSON <{}>".format(data))
         return data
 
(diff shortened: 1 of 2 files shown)


bogdanpetrea:
Hello @oucsaw, 
Thanks for this clean PR! The description should be saved when creating a subscription via the API.

Maybe the `description` and even the `reference` fields should be patchable, since these fields don't really interract with other mechanics of the Subscription / Billing parts.

LGTM

bogdanpetrea:
@oucsaw: Can you please rebase on top of master? We'll merge this after that.

oucsaw:
@bogdanpetrea  this should have been rebased to master.
Thanks!

PR #323, opened 2016-06-27
Allow api filtering by multiple references.

bogdanpetrea:
For example `/customers/1/subscriptions/?reference=1,2,3&state=ended`.


vtemian:
LGTM

PR #331, opened 2016-09-12
Split models (closes #257)

### LINKED ISSUE #257: Split the current models.py file in smaller files per class or groups of classes

bogdanpetrea:


silver/models/billing_entities.py
```diff
@@ -0,0 +1,215 @@
+# Copyright (c) 2016 Presslabs SRL
```
vtemian: Why not a module with all those entities in separate files?

silver/models/plans.py
```diff
@@ -0,0 +1,144 @@
...
+
+from django.core.exceptions import ValidationError
+from django.core.validators import MinValueValidator
+from django.db import models
+from django.utils.translation import ugettext_lazy as _
+
+from silver.utils.international import currencies
+
+
+class UnsavedForeignKey(models.ForeignKey):
```
vtemian: I would extract this in an utils module.

silver/models/documents.py
```diff
@@ -0,0 +1,706 @@
+# Copyright (c) 2016 Presslabs SRL
```
vtemian: Why not using a `documents` module which contains multiple files and each file a different model?

silver/models/subscriptions.py
```diff
@@ -0,0 +1,1242 @@
...
+                if field.name != 'metered_feature' and field.name != 'id':
+                    update_fields.append(field.name)
+            super(MeteredFeatureUnitsLog, self).save(
+                update_fields=update_fields)
+
+    def __unicode__(self):
+        return unicode(self.metered_feature.name)
+
+
+class Subscription(models.Model):
```
vtemian: Something is really wrong here. A single class containing around 1k lines of code smells.

bogdanpetrea: That's true. It's because document generation functionality was moved from it's own class into the Subscription class: https://github.com/PressLabs/silver/pull/158/commits/226d13af1b1d4c3f7675f63756d6d283fb79c463
https://github.com/PressLabs/silver/issues/94

silver/models/billing_entities.py
bogdanpetrea: I don't see a reason why not. :+1:

silver/models/documents.py
bogdanpetrea: :+1:

PR #335, opened 2016-09-27
Email changes

bogdanpetrea:
* Changed Customer `email` (EmailField) to `emails` (MultiEmailField). First email in this field will be used in templates.
* Changed Provider `email` to `display_email` (shown to customers) and added `internal_email` (used by Silver to communicate with the provider).


silver/models/billing_entities/customer.py
```diff
@@ -69,8 +72,20 @@ def get_archivable_field_values(self):
...
         customer_fields = ['customer_reference', 'consolidated_billing',
                            'payment_due_days', 'sales_tax_number',
-                           'sales_tax_percent']
+                           'sales_tax_percent', 'emails']
         fields_dict = {field: getattr(self, field, '') for field in
                        customer_fields}
         base_fields.update(fields_dict)
         return base_fields
+
+    def send_email(self, subject, body, from_email=None, cc=None, bcc=None):
```
vtemian: Send emails from models is not the best idea. The model layer should only be concern with data, not the business logic.

bogdanpetrea:
pong @vtemian

vtemian:
Why is this closed?

PR #337, opened 2016-11-11
Payment methods

bogdanpetrea:


silver/api/serializers.py
```diff
@@ -462,3 +462,27 @@ def update(self, instance, validated_data):
                 })
 
         return super(PaymentSerializer, self).update(instance, validated_data)
+
+
+class PaymentProcessorUrl(serializers.HyperlinkedRelatedField):
+    def __init__(self, view_name=None, **kwargs):
+        assert view_name is not None, 'The `view_name` argument is required.'
```
t-ionut: `HyperlinkedRelatedField` already takes care of this assertion in its own `__init__`. Isn't it safe to wait for the parent to crash 3 lines below?

silver/tests/spec/test_payment_processors.py
```diff
@@ -0,0 +1,55 @@
...
+
+        class SomeProcessor(GenericPaymentProcessor, TriggeredProcessorMixin):
+            name = "SomeProcessor"
+            @staticmethod
+            def setup(data=None):
+                pass
+
+        PaymentProcessorManager.register(SomeProcessor)
+
+        assert response.status_code == status.HTTP_200_OK
```
t-ionut: Shouldn't you have made another request before checking the response the second time?

silver/tests/spec/test_payment_processors.py
```diff
@@ -0,0 +1,55 @@
...
+        assert response.data == [OrderedDict([
+            ('name', u'Manual'),
+            ('type', u'manual'),
+            ('url', 'http://testserver/payment_processors/Manual/')
+        ])]
```
t-ionut: Shouldn't the response be expected to contain both the `ManualProcessor` and the `SomeProcessor` if you redo the request? Or am I reading this test wrong? :confused:

silver/models/transactions.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def __init__(self, *args, **kwargs):
+        self.form_class = kwargs.pop('form_class', None)
+
+        super(Transaction, self).__init__(*args, **kwargs)
+
+    @property
+    def is_usable(self):
+        return not self.disabled and \
+               not self.valid_until or \
```
t-ionut: So if the transaction is disabled but it isn't yet expired, then you can still use it?
If the transaction is disabled and there's no expiry date (valid_until is `None`) then I think you'll get a `TypeError` in your `self.valid_until > timezone.now()` comparison.

silver/api/serializers.py
```diff
@@ -465,24 +476,236 @@ def update(self, instance, validated_data):
...
-        view_name='payment-processor-detail', source='*',
+        view_name='payment-processor-detail', source='*', lookup_field="name",
+        read_only=True
+    )
+
+
+class PaymentMethodUrl(serializers.HyperlinkedRelatedField):
+    def get_url(self, obj, view_name, request, format):
+        kwargs = {'payment_method_id': obj.id,
+                  'customer_pk': obj.customer.id}
```
t-ionut: Maybe it doesn't matter in this specific case, but I think `pk` is the getter method meant to always give you the primary key, in case `id` is not the primary key.

http://stackoverflow.com/questions/2165865/django-queries-id-vs-pk

silver/tests/spec/test_payment_processors.py
krodyrobi: And shouldn't the new type be available in the ordered dict as well?

silver/api/serializers.py
```diff
@@ -465,24 +476,236 @@ def update(self, instance, validated_data):
...
+        return attrs
+
+    def validate_state(self, value):
+        if self.instance:
+            if value == self.instance.state:
+                return self.instance.state
+
+            message = "Illegal state transition from {} to {}."
+            if value == PaymentMethod.States.Uninitialized:
+                message.format(self.instance.state, value)
```
t-ionut:
```
>>> message = "This {}"
>>> message.format("won't work")
"This won't work"
>>> print(message)
This {}
```

silver/models/transactions.py
bogdanpetrea: :+1: I'll add some parentheses to that.

silver/models/payment_processors/generics.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def __str__(self):
+        return str(self.name)
+
+
+class PaymentProcessorTypes(object):
+    Manual = "manual"
+    Automatic = "automatic"
+    Triggered = "triggered"
+    Mixed = "mixed"
```
bogdanpetrea: Yeah, you're right, I'll remove that one. I'll update the wiki with a proper explanation soon.

silver/models/payment_processors/fields.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def from_db_value(self, value, *args, **kwargs):
+        return self.to_python(value)
+
+    def to_python(self, value):
+        if isinstance(value, GenericPaymentProcessor):
+            return value
+
+        if value is None:
+            return value
```
bogdanpetrea: True, but after the latest changes it won't anymore. "from_string_value" expects a string, not None.

silver/models/payment_processors/fields.py
```diff
@@ -0,0 +1,43 @@
...
+        max_length = 64
+        kwargs['max_length'] = max_length
+        super(PaymentProcessorField, self).__init__(*args, **kwargs)
+        self.validators = []
+
+    def get_internal_type(self):
+        return "CharField"
+
+    def from_string_value(self, value):
+        return PaymentProcessorManager.get(name=value) or value
```
bogdanpetrea: It returns value in case the payment_processor doesn't exist anymore in the PaymentProcessorManager.

silver/api/views.py
```diff
@@ -769,3 +773,60 @@ def put(self, request, *args, **kwargs):
...
+
+
+class PaymentProcessorDetail(RetrieveAPIView):
+    permission_classes = (permissions.IsAuthenticated,)
+    serializer_class = PaymentProcessorSerializer
+    ordering = ('-name', )
+
+    def get_object(self):
+        processor_name = self.kwargs.get('processor_name', '')
+        processor = PaymentProcessorManager.get(processor_name)
```
bogdanpetrea: :+1:

silver/api/views.py
```diff
@@ -769,3 +773,60 @@ def put(self, request, *args, **kwargs):
...
+        return get_object_or_404(Payment, customer__id=customer_pk,
+                                 pk=payment_pk)
+
+
+class PaymentProcessorList(ListAPIView):
+    permission_classes = (permissions.IsAuthenticated,)
+    serializer_class = PaymentProcessorSerializer
+    ordering = ('-name', )
+
+    def get_queryset(self):
```
bogdanpetrea: Yeah it can.

silver/api/serializers.py
bogdanpetrea: 👍 👍 👍

PR #367, opened 2016-11-17
Change payments api

t-ionut:
Fixes Presslabs/silver#361
Waiting for Presslabs/silver#364

### DIFF
--- a/silver/models/payments.py
+++ b/silver/models/payments.py
@@ -30,7 +30,6 @@ from django.dispatch import receiver
 from django.template.loader import select_template
 from django.utils.translation import ugettext_lazy as _
 
-from silver.models.transactions import Transaction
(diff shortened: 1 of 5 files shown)


vtemian:
@t-ionut this one needs a rebase. Also, it fails on travis.

silver/models/payments.py
```diff
@@ -154,6 +155,10 @@ def _log_unsuccessful_transition(self, transition_name):
         })
 
     @property
+    def transactions(self):
+        return Transaction.objects.filter(payment=self).all()
```
bogdanpetrea: You could use self.transaction_set.all() and avoid importing the Transaction class at all, as it might create some circular dependencies.

t-ionut:
@bogdanpetrea check again

PR #388, opened 2016-11-28
payment method tests fixes #343 

### LINKED ISSUE #343: Write tests for payment methods endpoints

krodyrobi:


krodyrobi:
I'll have a look over pytest for the transition test as it is hard too read as is. But just to facilitate faster integration a partial CR is needed.

silver/api/serializers.py
```diff
@@ -599,40 +600,18 @@ def validate_state(self, value):
...
-                                  PaymentMethod.States.Removed]):
-                message = message.format(self.instance.state, value)
-                raise serializers.ValidationError(message)
-
-            elif (self.instance.state == PaymentMethod.States.Disabled and
-                    value not in [PaymentMethod.States.Enabled,
-                                  PaymentMethod.States.Removed]):
-                message = message.format(self.instance.state, value)
-                raise serializers.ValidationError(message)
+        if value not in [x[0] for x in PaymentMethod.States.Choices]:
```
vtemian: Can you replace `value` with `state` and `x` with `choice`?

silver/api/serializers.py
```diff
@@ -599,40 +600,18 @@ def validate_state(self, value):
...
+            raise serializers.ValidationError('Unknown state {}'.format(value))
 
         return value
 
-    def validate_customer(self, value):
```
vtemian: Don't we need this one?

silver/api/serializers.py
```diff
@@ -660,40 +639,38 @@ def create(self, validated_data):
         return payment_method
 
     def update(self, instance, validated_data):
-        state = validated_data.pop('state', None)
+        old_state = instance.state
```
vtemian: Can you refactor this method?
Is very hard to read and understand

silver/tests/spec/test_payment_method.py
```diff
@@ -1,49 +1,330 @@
...
+        super(TestPaymentMethodEndpoints, self).setUp()
+
+    def create_payment_method(self, *args, **kwargs):
+        method = PaymentMethodFactory.create(*args, **kwargs)
 
         # mysql does not store fractional time units but the object
         # created will have them so we can't use it directly
         #  to check the output
-        payment_method.refresh_from_db()
+        method.refresh_from_db()
```
vtemian: Why do you need to refresh it if you just created it?

silver/tests/spec/test_payment_method.py
```diff
@@ -1,49 +1,330 @@
...
+
+        return method
+
+    def test_get_listing(self):
+        PaymentMethodFactory.create(customer=CustomerFactory.create())
```
vtemian: Why do you need this one?

silver/tests/spec/test_payment_method.py
```diff
@@ -1,49 +1,330 @@
...
+        response = self.client.post(url, data={
+            'payment_processor': processor_url,
+            'state': PaymentMethod.States.Uninitialized,
+            'additional_data': '{"random": "value"}'
+        }, format='json')
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+    def test_post_listing_invalid_initial_state(self):
+        invalid_initial_states = list(
```
vtemian:
```python
invalid_initial_states = [state[0] for state in PaymentMethod.States.Choices
                                        if state[0] not in PaymentMethod.InitialStates]
```

silver/tests/spec/test_payment_method.py
```diff
@@ -1,49 +1,330 @@
...
+        for state in invalid_initial_states:
+            response = self.client.post(url, data={
+                'payment_processor': processor_url,
+                'state': state
+            }, format='json')
+
+            self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+    def test_put_detail_state_transitions(self):
+        states = [x[0] for x in PaymentMethod.States.Choices]
```
vtemian: Why not storing them in `PaymentMethod.States`?

silver/tests/spec/test_payment_method.py
```diff
@@ -1,49 +1,330 @@
...
+        permutations = [(old, new) for old in states for new in states]
```
vtemian: You can use `permutations` from itertools https://docs.python.org/2/library/itertools.html#itertools.permutations

silver/tests/spec/test_payment_method.py
```diff
@@ -1,49 +1,330 @@
...
+
+        valid_transitions = set()
+        for _, transition in iteritems(PaymentMethod.registered_transitions):
```
vtemian: Why not PaymentMethod.registered_transition.iteritems() ?

silver/tests/spec/test_payment_method.py
krodyrobi: The comment above. When creating and serializing vs api serialization are 2 different things.
If I serialize the object directly it will have a datetime with microseconds if I get it from the db it won't have that data and won't be identical.

silver/tests/spec/test_payment_method.py
krodyrobi: So I see if querying on a specific client limits the returned data to it.
2 Clients each with 1 payment method. I'm expecting for a url of form /customer/1/payment_methods to receive only those linked with customer 1.

silver/tests/spec/test_payment_method.py
krodyrobi: No need for another import just for that one line alone.
And yes there is even a better solution than permutations i need the `itertools.product`

silver/tests/spec/test_payment_method.py
krodyrobi: six compatibility for future thinking.

silver/tests/spec/test_payment_method.py
krodyrobi: And honestly parsing datetime to convert and stuff would end up in more lines of code, One solution would be to force django to store microseconds in mysql but then we would need to check if mysql can take that route as not all versions can...

silver/tests/spec/test_payment_method.py
bogdanpetrea: Here's how I did it for transactions: https://github.com/PressLabs/silver/pull/389/files#diff-2ba583c21bc7117462ac341cfc3d734fR28

silver/tests/spec/test_payment_method.py
```diff
@@ -1,49 +1,330 @@
...
+            if isinstance(transition['source'], (tuple, list)):
+                unmerged_transitions = set([(old, transition['target']) for
+                                            old in transition['source']])
+            else:
+                unmerged_transitions = {(transition['source'],
+                                         transition['target'])}
+
+            valid_transitions = valid_transitions.union(unmerged_transitions)
+
+        # add x -> x transactions as they are allowed
```
bogdanpetrea: I think you meant transitions here right? And no, they're not allowed.

silver/tests/spec/test_payment_method.py
```diff
@@ -1,49 +1,330 @@
...
+                                            state=PaymentMethod.States.Disabled)
+
+        url = reverse('payment-method-detail', kwargs={
+            'customer_pk': self.customer.pk,
+            'payment_method_id': method.pk
+        })
+        response = self.client.get(url, format='json')
+
+        data = response.data
+        data['additional_data'] = '{"random":"value"}'
```
bogdanpetrea: Missing whitespace after `:`.

silver/tests/spec/test_payment_method.py
```diff
@@ -1,49 +1,330 @@
...
+        data = response.data
+        data['customer'] = reverse('customer-detail',
+                                   request=response.wsgi_request,
+                                   kwargs={'pk': other_customer.pk})
+
+        response = self.client.put(url, data=data, format='json')
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data, expected_data)
+
+    @register
```
bogdanpetrea: Maybe move this to a common place and maybe rename it to register_processor or something more meaningful.

silver/tests/spec/test_payment_method.py
krodyrobi: That is what I meant. But in theory nothing stops you from doing it except that we short-circuit the logic if we detect no change, but technically still doable.

krodyrobi:
@bogdanpetrea  @vtemian  @t-ionut last call.

silver/api/serializers.py
```diff
@@ -432,7 +441,8 @@ def validate(self, attrs):
             )
             raise serializers.ValidationError(message)
         if (self.instance and getattr(self.instance, 'transactions', None) and
-           self.instance.transaction_set.exclude(state='canceled').exists()):
+                self.instance.transaction_set.exclude(
+                    state='canceled').exists()):
```
bogdanpetrea: ?

silver/api/serializers.py
```diff
@@ -460,7 +470,7 @@ def validate(self, attrs):
             if non_field_errors:
                 errors['non_field_errors'] = [
                     error for sublist in non_field_errors for error in sublist
-                ]
+                    ]
```
bogdanpetrea: ?

silver/api/serializers.py
```diff
@@ -575,7 +587,7 @@ def validate(self, attrs):
         # Update
         else:
             if (additional_data and
-                    self.instance.state == PaymentMethod.States.Disabled):
+                        self.instance.state == PaymentMethod.States.Disabled):
```
bogdanpetrea: ?

silver/api/serializers.py
```diff
@@ -588,51 +600,29 @@ def validate(self, attrs):
             if state not in allowed_initialized_states:
                 message = "If 'additional_data' is specified, " \
                           "then 'state' must be one of ({}).".format(
-                              ', '.join(allowed_initialized_states)
-                          )
+                    ', '.join(allowed_initialized_states)
```
bogdanpetrea: ?

silver/models/payment_methods.py
```diff
@@ -41,7 +41,37 @@ class States(object):
...
+            return [choice[0] for choice in cls.Choices]
+
     state = FSMField(choices=States.Choices, default=States.Uninitialized)
+    state_transitions = {
+        'initialize_unverified': {
+            'source': States.Uninitialized,
+            'target': States.Unverified
+        },
+        'initialize_enabled': {
+            'source': States.Unverified,
```
bogdanpetrea: States.Unitialized

silver/tests/spec/test_payment_method.py
```diff
@@ -1,27 +1,308 @@
...
+            'customer_pk': self.customer.pk
+        })
+
+        response = self.client.post(url, data={
+            'payment_processor': processor_url,
+            'state': PaymentMethod.States.Uninitialized,
+            'additional_data': '{"random": "value"}'
+        }, format='json')
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
```
bogdanpetrea: It would be nice to test `response.data` as well.

silver/tests/spec/test_payment_method.py
```diff
@@ -1,27 +1,308 @@
...
+                                 message.format(response.status_code,
+                                                status.HTTP_200_OK))
+                self.assertEqual(data, response.data,
+                                 message.format(response.status_code,
+                                                status.HTTP_200_OK))
+            else:
+                self.assertEqual(response.status_code,
+                                 status.HTTP_409_CONFLICT,
+                                 message.format(response.status_code,
+                                                status.HTTP_409_CONFLICT))
```
bogdanpetrea: We might need a test for this test. 😁

silver/tests/spec/test_payment_method.py
```diff
@@ -1,27 +1,308 @@
...
+
+        self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
+
+    def test_post_listing_incomplete_body_1(self):
+        url = reverse('payment-method-list', kwargs={
+            'customer_pk': 0
+        })
+
+        response = self.client.post(url, data={}, format='json')
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
```
bogdanpetrea: Here as well. `response.data` Probably elsewhere too, but won't spam the review with this.

silver/tests/spec/util/api_get_assert.py
```diff
@@ -12,11 +12,12 @@ def setUp(self):
         self.client.force_authenticate(user=admin_user)
 
     def assert_get_data(self, url, expected_data):
+        many = isinstance(expected_data, list)
         response = self.client.get(url, format='json')
         request = response.wsgi_request
         expected = self.serializer_class(expected_data,
```
bogdanpetrea: I don't think we should use the serializer to test views that output results based on the same serializer. Get my point?

silver/api/serializers.py
krodyrobi: Pycharm

silver/api/serializers.py
krodyrobi: Pycharm

silver/api/serializers.py
krodyrobi: Don't even

silver/tests/spec/test_payment_method.py
krodyrobi: So I'm stuck with pytest then. So be it.

silver/tests/spec/util/api_get_assert.py
krodyrobi: I'll create a serializer encoding test to make sure it doesn't change without our knowledge.

silver/tests/spec/test_payment_method.py
krodyrobi: There is no point for this as of yet as the code that should generate the lattice of transitions should still be there, we'll only loose time but not gain much in return (maybe more verbosity for successfully ran tests, not worth it imho).

PR #389, opened 2016-11-29
Merge payments with transactions

bogdanpetrea:


silver/models/transactions.py
```diff
@@ -1,31 +1,219 @@
...
+                         send_refunded_transaction_email,
+                         send_settled_transaction_email)
+
+from silver.utils.international import currencies
+
+
+logger = logging.getLogger(__name__)
+
+
+class _States(type):
```
vtemian: Not fun :)

silver/models/transactions.py
```diff
@@ -1,31 +1,219 @@
...
+
+    @property
     def payment_processor(self):
         return self.payment_method.payment_processor
+
+    def __unicode__(self):
+        return unicode(self.uuid)
+
+
+def send_email_with_current_state(transaction):
```
vtemian: You can use a dict and as values to have those functions:

```python
actions = {
    Transaction.States.Initial: send_new_transaction_email,
    ...
}
return actions[transaction.state](transaction)
if transaction.state in actions else None
```

silver/models/transactions.py
```diff
@@ -1,31 +1,219 @@
...
+            send_email_with_current_state(instance)
+
+
+@receiver(post_transition)
+def post_transition_callback(sender, instance, name, source, target, **kwargs):
+    """
+    Syncs the state of the related documents of the transaction with the
+    transaction state
+    """
+    if issubclass(sender, Transaction):
```
vtemian: You can negate this and the code should look a little bit better.

silver/models/transactions.py
```diff
@@ -1,31 +1,219 @@
...
+        if target == Transaction.States.Settled:
```
vtemian: The same here.
Another idea would be to extract this logic in another function, in case that the decision branches will grow.

silver/tests/unit/test_invoice.py
```diff
@@ -140,3 +143,94 @@ def test_issues_invoice_series_number(self):
...
+        invoices[2].due_date = date.today() - timedelta(days=31)
+        invoices[2].save()
+
+        invoices[3].state = Invoice.STATES.CANCELED
+        invoices[3].save()
+
+        queryset = Invoice.objects.due_this_month()
+
+        assert queryset.count() == 2
+        for payment in invoices[0:1]:
```
vtemian: This is the same with `for payment in invoices[0]`

silver/tests/unit/test_invoice.py
```diff
@@ -140,3 +143,94 @@ def test_issues_invoice_series_number(self):
...
+        invoices[1].save()
+
+        invoices[2].state = Invoice.STATES.PAID
+        invoices[2].due_date = date.today() - timedelta(days=31)
+        invoices[2].save()
+
+        queryset = Invoice.objects.overdue()
+
+        assert queryset.count() == 2
+        for payment in invoices[0:1]:
```
vtemian: Same as above.

silver/models/transactions.py
bogdanpetrea: Well, blame python for that one.

silver/models/transactions.py
```diff
@@ -1,31 +1,219 @@
...
+
+@receiver(pre_save, sender=Transaction)
+def pre_transaction_save(sender, instance=None, **kwargs):
+    old = get_object_or_None(Transaction, pk=instance.pk)
+    setattr(instance, 'old_value', old)
+
+
+@receiver(post_save, sender=Transaction)
+def post_transaction_save(sender, instance, **kwargs):
+    if not instance.old_value:
```
krodyrobi: You can negate here as well.

silver/models/transactions.py
bogdanpetrea: It makes sense to negate the first if, because this callback is only for Transactions (a hard requirement), but not so much for the target checking.

You are right about wrapping the logic in functions when the branches will grow, but for now 2 decision branches seem fair.

PR #399, opened 2016-12-27
Further work on Braintree.

bogdanpetrea:


payment_processors/braintree_processor/models/payment_processors.py
```diff
@@ -94,18 +89,17 @@ def _update_transaction_status(self, transaction, result_transaction):
...
 
-        transaction.data.update({
-            'status': result_transaction.status,
-            'braintree_id': result_transaction.id
-        })
+        transaction.external_reference = result_transaction.id
+        status = result_transaction.status
 
-        status = transaction.data['status']
+        transaction.data['status'] = status
```
bogdanpetrea: Because I used it a few lines below too, and didn't want to write `result_transaction.status` all the time.

payment_processors/braintree_processor/models/payment_processors.py
```diff
@@ -6,12 +6,11 @@
 from django.utils import timezone
 from django_fsm import TransitionNotAllowed
 
+from .payment_methods import BraintreePaymentMethod
+from ..views import BraintreeTransactionView
```
vtemian: Not fun. Maybe use absolut imports.

payment_processors/braintree_processor/models/payment_processors.py
```diff
@@ -44,43 +43,39 @@ def refund_transaction(self, transaction, payment_method=None):
...
-            'type': result_payment_method.__class__.__name__
+            'type': type,
+            'image_url': result_details.image_url,
+            'updated_at': timezone.now().isoformat()
         }
 
-        if payment_method_details['type'] == payment_method.Type.PayPal:
-            payment_method_details['email'] = result_payment_method.email
-        elif payment_method_details['type'] == payment_method.Type.CreditCard:
+        if type == payment_method.Types.PayPal:
```
vtemian: `type` is a built-in function in Python. Use `kind`,  another synonym or `payment_type`.

payment_processors/braintree_processor/models/payment_processors.py
```diff
@@ -160,21 +151,46 @@ def _charge_transaction(self, transaction):
...
         # send transaction request
         result = braintree.Transaction.sale(data)
 
         # handle response
         if result.is_success and result.transaction:
-            self._update_payment_method(payment_method,
-                                        result.transaction.payment_method)
+            self._update_customer(customer, result.transaction.customer_details)
+
+            type = result.transaction.payment_instrument_type
```
vtemian: Same as above.

payment_processors/braintree_processor/models/payment_processors.py
bogdanpetrea: Well I don't think it's that bad. In Django projects, the views are usually on the same level with the models, but since in this case models is a package, I just had to go a level up.

vtemian: http://stackoverflow.com/questions/4209641/absolute-vs-explicit-relative-import-of-python-module

bogdanpetrea: The accepted answer from there is from 2010 and I disagree with it.

PR #400, opened 2016-12-28
Added a manage_transactions command.

bogdanpetrea:


silver/management/commands/manage_transactions.py
```diff
@@ -0,0 +1,62 @@
...
+        if options['transactions']:
+            manageable_transactions = manageable_transactions.filter(
+                transaction__pk__in=options['transactions']
+            )
+
+        for transaction in manageable_transactions:
+            try:
+                transaction.processor.manage_transaction(transaction)
+            except Exception as e:
+                logger.info('Encountered exception while managing transaction '
```
vtemian: ```logger.error(<your message here>, exc_info=True)``` it would be

PR #414, opened 2017-01-03
Refactor payment processors rendering methods (fixes #406)

### LINKED ISSUE #406: Rework PaymentProcessor form rendering
Currently it looks like this:
```
class PaymentProcessor
    transaction_view_class = ViewClass

class TransactionViewClass
    form_class = FormClass
```

After rework it should like this
```
class PaymentProcessor
    transaction_view_class = ViewClass
    form_class = FormClass

    get_view()
         return transaction_view_class(get_form(), get_template() ...)
    get_form()
         return form_class()
    get_template()
         return template(paymentprocessor.reference, provider) or
                   template(paymentprocessor.reference) or
                   template(default)
```

vtemian:


silver/models/payment_processors/base.py
```diff
@@ -13,6 +15,36 @@ def setup(self, data):
...
+    def get_form(self, transaction, request):
+        form = None
+        if self.form_class:
+            form = self.form_class(payment_method=transaction.payment_method,
+                                   transaction=transaction, request=request)
+
+        return form
+
+    def get_template(self, transaction):
+        template = select_template([
```
t-ionut: Why move the template selection logic into the "model"? AFAICS Django itself leaves this to the view/s

vtemian: Because I want to abstract the way in which a Payment Processor will chose which template to render, for `pay_url` view.
Each new payment processor will have to inherit from this base and if that will want to enforce a specific, custom one, it can by overriding this method. Otherwise, it will choose from those 3 standard paths.

PR #415, opened 2017-01-03
Remove mail notifications. (fixes #410)

### LINKED ISSUE #410: Remove mailing from silver

vtemian:


vtemian:
Silver offers signals and any implementation of silver should choose what kind of notification to use for each specific silver action (payed invoices/proforma, different state transitions etc.)

PR #416, opened 2017-01-04
Replace manage_transaction with execute_transaction

t-ionut:
Fixes https://github.com/PressLabs/silver/issues/409
Other than renaming everywhere, I don't see any logic that needs to be moved. The Braintree processor currently changes the transaction state in its own implementation, which is only called from `execute_transaction`.
CR please and let me know if I went about this wrong.


bogdanpetrea:
LGTM

PR #422, opened 2017-01-06
Refactor payment processors (fixes #408)

### LINKED ISSUE #408: Remove PaymentMethod states
The current states can be seen here: https://github.com/PressLabs/silver/wiki/Resources#paymentmethodstate

To replace the states the following fields should be implemented:
`verified = BooleanField(default=False)` - If a payment method is verified it means it can be trusted to make payments (e.g. It was verified by the payment processor service or a transaction has been settled using it)

`disabled = BooleanField(default=False)` - Indicates a payment method is no longer usable.

Progress
- [x] Model changes
- [ ] Project changes
- [ ] Wiki updated

vtemian:


silver/tests/spec/test_payment_method.py
```diff
@@ -353,16 +261,8 @@ def test_filter_processor(self):
...
-            'customer_pk': self.customer.pk
-        })
-
-        url_state_enabled = url + '?state=' + PaymentMethod.States.Enabled
-        url_no_output = url + '?state=' + PaymentMethod.States.Disabled
-
-        self.assert_get_data(url_state_enabled, [payment_method])
-        self.assert_get_data(url_no_output, [])
+    def test_filter_verified(self):
+        pass
```
bogdanpetrea: You shall not pass!

PR #460, opened 2017-01-17
Handle redirect logic in silver.

vtemian:
Also `pay_url` has an expiration time. (fixes https://github.com/silverapp/silver/issues/458)

### DIFF
--- a/silver/api/serializers.py
+++ b/silver/api/serializers.py
@@ -493,10 +493,12 @@ class TransactionPaymentUrl(serializers.HyperlinkedIdentityField):
         return get_payment_url(obj, request)
 
     def get_object(self, view_name, view_args, view_kwargs):
-        transaction_uuid = jwt.decode(view_kwargs['token'],
-                                      settings.PAYMENT_METHOD_SECRET,
-                                      verfify=False)['transaction']
-        return self.queryset.get(uuid=transaction_uuid)
+        try:
+            transaction_uuid = jwt.decode(view_kwargs['token'],
+                                          settings.PAYMENT_METHOD_SECRET)['transaction']
+            return self.queryset.get(uuid=transaction_uuid)
+        except (jwt.ExpiredSignatureError, jwt.DecodeError, jwt.InvalidTokenError):
+            return None
 
 
 class TransactionSerializer(serializers.HyperlinkedModelSerializer):
--- a/silver/api/serializers.py
+++ b/silver/api/serializers.py
@@ -35,6 +35,7 @@ from silver.models import (MeteredFeatureUnitsLog, Customer, Subscription,
                            DocumentEntry, ProductCode, Proforma, PaymentMethod,
                            PaymentProcessorManager, Transaction)
 from silver.validators import validate_payment_processor
+from silver.utils.payments import get_payment_url
 
 
 class ProductCodeRelatedField(serializers.SlugRelatedField):
@@ -489,12 +490,7 @@ class PaymentMethodSerializer(serializers.HyperlinkedModelSerializer):
 
 class TransactionPaymentUrl(serializers.HyperlinkedIdentityField):
     def get_url(self, obj, view_name, request, format):
-        lookup_value = jwt.encode({
-            'exp': datetime.datetime.utcnow() + datetime.timedelta(seconds=30),
-            'transaction': str(getattr(obj, self.lookup_field))
-        }, settings.PAYMENT_METHOD_SECRET)
-        kwargs = {'token': str(lookup_value)}
-        return self.reverse(view_name, kwargs=kwargs, request=request, format=format)
+        return get_payment_url(obj, request)
 
     def get_object(self, view_name, view_args, view_kwargs):
         transaction_uuid = jwt.decode(view_kwargs['token'],
--- a/silver/api/serializers.py
+++ b/silver/api/serializers.py
@@ -504,7 +504,7 @@ class TransactionSerializer(serializers.HyperlinkedModelSerializer):
         fields = ('id', 'url', 'customer', 'provider', 'amount', 'currency',
                   'currency_rate_date', 'state', 'proforma', 'invoice',
                   'can_be_consumed', 'payment_processor', 'payment_method',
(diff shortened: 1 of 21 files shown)


silver/models/payment_processors/base.py
```diff
@@ -53,6 +53,9 @@ def get_template(self, transaction):
 
         return template
 
+    def was_transaction_initialized(self, transaction, request):
```
bogdanpetrea: Maybe it would make sense, to rename this to `handle_transaction_response` or something similar (maybe outcome or feedback instead of response) and have it only update the transaction status based on the request (either made by the client or the payment service provider).

Also a comment in the code, about what the method should do would be nice.

silver/views.py
```diff
@@ -43,6 +43,26 @@ def invoice_pdf(request, invoice_id):
...
+        uuid = UUID(transaction_uuid, version=4)
+    except ValueError:
+        raise Http404
+
+    transaction = get_object_or_404(Transaction, uuid=uuid)
+
+    transaction.last_access = timezone.now()
+    transaction.save()
+
+    if transaction.payment_processor.was_transaction_initialized(transaction,
```
bogdanpetrea: And here you'd only do `handle_transaction_response(transaction, request)` and based on the `transaction.status` you redirect to the proper url. This way we can have a `canceled_url` too, which is pretty common to have.

silver/views.py
```diff
@@ -43,6 +43,26 @@ def invoice_pdf(request, invoice_id):
 
 
 @csrf_exempt
+def initialize_transaction(request, transaction_uuid):
```
bogdanpetrea: Rename this too.

silver/models/payment_processors/base.py
```diff
@@ -53,6 +53,9 @@ def get_template(self, transaction):
 
         return template
 
+    def handle_transaction_response(self, transaction, request):
```
bogdanpetrea: Comment about what this method should do.

silver/utils/decorators.py
```diff
@@ -0,0 +1,43 @@
...
+            expired = True
+            transaction_uuid = jwt.decode(token, settings.PAYMENT_METHOD_SECRET,
+                                          options={'verify_exp': False})['transaction']
+
+        try:
+            uuid = UUID(transaction_uuid, version=4)
+        except ValueError:
+            raise NotFound
+
+        return view(request, get_object_or_404(Transaction, uuid=uuid), expired)
```
bogdanpetrea: 👍

silver/utils/payments.py
```diff
@@ -34,4 +35,9 @@ def get_payment_url(transaction, request):
 
 def get_payment_complete_url(transaction, request):
     kwargs = {'token': str(_get_jwt_token(transaction))}
-    return reverse('payment-complete', kwargs=kwargs, request=request)
+    url = furl(reverse('payment-complete', kwargs=kwargs, request=request))
```
bogdanpetrea: :+1:

silver/views.py
```diff
@@ -50,9 +51,9 @@ def complete_payment_view(request, transaction, expired=None):
         transaction.save()
 
     if 'return_url' in request.GET:
-        # TODO: use urlparse, as utils
-        redirect_url = "{}?transaction_uuid={}".format(request.GET['return_url'],
-                                                       transaction.uuid)
+        redirect_url = furl(request.GET['return_url']).add({
```
bogdanpetrea: :+1:

silver/api/serializers.py
```diff
@@ -483,14 +488,24 @@ def validate_payment_processor(self, value):
         return value
 
 
+class TransactionPaymentUrl(serializers.HyperlinkedIdentityField):
+    def get_url(self, obj, view_name, request, format):
+        return get_payment_url(obj, request)
+
+    def get_object(self, view_name, view_args, view_kwargs):
+        transaction_uuid = jwt.decode(view_kwargs['token'],
```
vtemian: I was thinking the same, but maybe one day somebody will use it.

silver/tests/unit/test_payments_util.py
```diff
@@ -0,0 +1,48 @@
...
+# limitations under the License.
+from mock import patch, MagicMock
+
+from django.test import TestCase
+
+from silver.tests.factories import TransactionFactory
+from silver.utils.payments import get_payment_url, get_payment_complete_url
+
+
+class TestPaymentsUtilMethods(TestCase):
```
bogdanpetrea: I think a couple of tests for `_get_jwt_token` is the crucial thing here.

silver/views.py
```diff
@@ -67,7 +67,7 @@ def complete_payment_view(request, transaction, expired=None):
 @get_transaction_from_token
 def pay_transaction_view(request, transaction, expired=None):
     if expired:
-        raise NotFound
+        raise Http404
```
bogdanpetrea: 👍

silver/tests/spec/test_payments.py
```diff
@@ -0,0 +1,70 @@
...
+        self.user = AdminUserFactory.create()
+        self.client.force_authenticate(user=self.user)
+
+    def test_pay_transaction_view_expired(self):
+        customer = CustomerFactory.create()
+        payment_method = PaymentMethodFactory.create(customer=customer)
+        transaction = TransactionFactory.create(payment_method=payment_method)
+
+        with patch('silver.utils.payments.datetime') as mocked_datetime:
+            mocked_datetime.utcnow.return_value = datetime.utcnow() - timedelta(days=365)
```
bogdanpetrea: Maybe leave a comment, explaining the magic: this will create an expired token such that the view will return 404. Also renaming the test to `test_pay_transaction_view_expired_token` might offer more insight about what this test does.

silver/models/payment_processors/base.py
```diff
@@ -54,6 +54,17 @@ def get_template(self, transaction):
         return template
 
     def handle_transaction_response(self, transaction, request):
+        """
```
bogdanpetrea: :+1:

PR #462, opened 2017-01-20
More currencies

bogdanpetrea:


silver/admin.py
```diff
@@ -257,8 +259,8 @@ def end(self, request, queryset):
 
 class CustomerAdmin(LiveModelAdmin):
     fields = ['company', 'first_name', 'last_name', 'customer_reference',
-              'email', 'address_1', 'address_2', 'city', 'state', 'zip_code',
-              'country', 'consolidated_billing', 'payment_due_days',
+              'emails', 'address_1', 'address_2', 'city', 'state', 'zip_code',
```
vtemian: Here sould be `email`

silver/admin.py
```diff
@@ -468,6 +474,21 @@ def __init__(self, *args, **kwargs):
 
         super(BillingDocumentForm, self).__init__(*args, **kwargs)
 
+    def clean(self, *args, **kwargs):
+        cleaned_data = super(BillingDocumentForm, self).clean(*args, **kwargs)
+
+        customer = cleaned_data['customer']
+        currency = cleaned_data['currency']
+
+        cleaned_data['transaction_currency'] = (
```
vtemian: Maybe add a validation for `transaction_currency` to be a value from `currencies`

silver/models/currency/base.py
```diff
@@ -0,0 +1,14 @@
+from .exceptions import RateNotFound
+
+
+class CurrencyConverterBase(object):
```
vtemian: What about a method called `get_xe_rate(to_currency, from_currency, date)?`

silver/models/documents/base.py
```diff
@@ -149,6 +165,19 @@ def _issue(self, issue_date=None, due_date=None):
...
+        if not self.transaction_xe_rate:
+            if not self.transaction_xe_date:
+                self.transaction_xe_date = self.issue_date - timedelta(days=1)
+
+            try:
+                xe_rate = CurrencyConverter.convert(1, self.currency,
+                                                    self.transaction_currency,
+                                                    self.transaction_xe_date)
+            except RateNotFound:
+                raise TransitionNotAllowed
```
vtemian: Maybe a message would help to find out why the transition is not allowed

silver/models/documents/base.py
```diff
@@ -343,6 +375,11 @@ def _entries(self):
     def transactions(self):
         return self.transaction_set.all()
 
+    @property
+    def transaction_total(self):
+        return (self.total * self.transaction_xe_rate if self.transaction_xe_rate
```
vtemian: Why not considering `transaction_xe_rate` to have 1 as default.
```
return (self.total * self.transaction_xe_rate if self.transaction_xe_rate else self.total)
```

silver/models/payment_methods.py
```diff
@@ -43,12 +43,6 @@ class PaymentMethod(models.Model):
 
     objects = InheritanceManager()
 
-    customer = models.ForeignKey(Customer)
-    added_at = models.DateTimeField(default=timezone.now)
-    data = JSONField(blank=True, null=True, default={})
-
-    objects = InheritanceManager()
```
vtemian: :+1:

silver/admin.py
bogdanpetrea: There's a `clean_currency` in the Transaction model.

silver/models/currency/base.py
bogdanpetrea: That's just `convert(1, to_currency, from_currency, date)`.

silver/models/documents/base.py
bogdanpetrea: Because it could leave to currency errors. `transaction_xe_rate` always exists if the document was `issued` and you can only create transactions for `issued` documents so theoretically you don't need `transaction_total` when `transaction_xe_rate` is not set.

PR #465, opened 2017-01-22
If the transaction can't be paid, don't show it's pay_url. (fixes #435)

### LINKED ISSUE #435: Don't show `pay_url` if the transaction can't be paid
eg: the state is not Initial or the transaction is expired.

Raise 404 for a transaction that can't be paid https://github.com/silverapp/silver/blob/63a53114dc5edb36ef902c32810a031e2c7219bb/silver/views.py#L56

vtemian:


silver/api/serializers.py
```diff
@@ -490,6 +490,8 @@ def validate_payment_processor(self, value):
 
 class TransactionPaymentUrl(serializers.HyperlinkedIdentityField):
     def get_url(self, obj, view_name, request, format):
+        if not obj.can_be_consumed:
+            return ''
```
bogdanpetrea: I think it's better to return `None` here.

PR #468, opened 2017-01-23
Added update_at and created_at to api.

vtemian:


silver/api/serializers.py
```diff
@@ -524,7 +524,7 @@ class Meta:
         fields = ('id', 'url', 'customer', 'provider', 'amount', 'currency',
                   'currency_rate_date', 'state', 'proforma', 'invoice',
                   'can_be_consumed', 'payment_processor', 'payment_method',
-                  'pay_url', 'valid_until')
+                  'pay_url', 'valid_until', 'updated_at', 'created_at')
```
bogdanpetrea: I think these might need to be declared as `read-only` too.

PR #479, opened 2017-01-26
Moved payment_processors out of models

calind:
Moved `payment_processors` out of `silver.models` because they are not models. Also now the imports make more sense, like `from silver.payment_processors import PaymentProcessorBase`. 

I started this because there is no easy way to load fixtures. `PaymentProcessorManager` is a bit more magical than required, and the value serialized in fixtures cannot be unpacked. We should have dropped it a long time ago. Also dropped the `PaymentProcessorField` since it's not required any more.

I'm considering renaming `PaymentProcessorBase` to `PaymentProcessor` since it's more succinct.

ToDo:
- [ ] update remaining tests
- [ ] update `silver-payu` and `silver-braintree`


vtemian:
I don't think that this is really necessary right now.
I would focus on launching a beta version this week and then improve it.

PR #491, opened 2017-01-30
Keep invoice/proforma in sync on transaction updates

calind:


silver/models/transactions.py
```diff
@@ -235,6 +240,11 @@ def pre_transaction_save(sender, instance=None, **kwargs):
     if not getattr(instance, 'cleaned', False):
         instance.full_clean()
 
+    transaction = instance
```
vtemian: You can move this one as the first line in this function.

silver/models/transactions.py
```diff
@@ -235,6 +240,11 @@ def pre_transaction_save(sender, instance=None, **kwargs):
...
+    proforma = transaction.proforma
+    if proforma and proforma.invoice and transaction.invoice is None:
```
vtemian: `if proforma and proforma.invoice and not transaction.invoice:`

silver/tests/factories.py
```diff
@@ -247,6 +247,23 @@ class Meta:
     customer = factory.SubFactory(CustomerFactory)
 
 
+class ProformaTransactionFactory(factory.django.DjangoModelFactory):
```
vtemian: `TransactionProformaFactory`

bogdanpetrea: Maybe use `TransactionFactory` like this: `TransactionFactory.create(invoice=None)`. Might need to make it work first.

Issue #486, opened 2017-01-30
Disable payment method with side effects


vtemian:
Add a `disable` method to `PaymentMethod` that will put `enable` to `False` and move all transactions to `canceled`.

PR #493, opened 2017-01-30
Disable payment method (fixes #487 and #486)

### LINKED ISSUE #487: Rename `enabled` to `canceled` and make it a `read-only` field

vtemian:


silver/api/urls.py
```diff
@@ -41,6 +41,8 @@
         views.PaymentMethodList.as_view(), name='payment-method-list'),
     url(r'^customers/(?P<customer_pk>[0-9]+)/payment_methods/(?P<payment_method_id>[0-9]+)/$',
         views.PaymentMethodDetail.as_view(), name='payment-method-detail'),
+    url(r'^customers/(?P<customer_pk>[0-9]+)/payment_methods/(?P<payment_method_id>[0-9]+)/(?P<requested_action>[\w]+)_request/$',
+        views.PaymentMethodDetail.as_view(), name='payment-method-detail'),
```
bogdanpetrea: I think you meant `PaymentMethodAction` here. Use a different view-name as well.

silver/api/urls.py
```diff
@@ -41,6 +41,8 @@
         views.PaymentMethodList.as_view(), name='payment-method-list'),
     url(r'^customers/(?P<customer_pk>[0-9]+)/payment_methods/(?P<payment_method_id>[0-9]+)/$',
         views.PaymentMethodDetail.as_view(), name='payment-method-detail'),
+    url(r'^customers/(?P<customer_pk>[0-9]+)/payment_methods/(?P<payment_method_id>[0-9]+)/(?P<requested_action>[\w]+)_request/$',
```
bogdanpetrea: I suggest we either hardcode every possible action (maybe with a regex) and / or validate `requested_action` in `PaymentMethodAction`.

silver/api/views.py
```diff
@@ -863,6 +863,30 @@ def get_object(self):
...
+            raise Http404
+
+        action_to_execute()
+
+    def get_object(self, **kwargs):
+        payment_method_id = kwargs.get('payment_method_id')
+        customer_pk = kwargs.get('customer_pk')
+
+        return get_object_or_404(
+            PaymentMethod.objects.all().select_subclasses(),
```
bogdanpetrea: 👍

silver/models/payment_methods.py
```diff
@@ -84,6 +89,35 @@ def decrypt_data(self, crypted_data):
...
+                             Transaction.States.Pending]
+        transactions = Transaction.objects.filter(payment_method=self,
+                                                  state__in=cancelable_states)
+
+        for transaction in transactions:
+            if transaction.state == Transaction.States.Initial:
+                transaction.cancel()
+
+            if transaction.state == Transaction.States.Pending:
+                transaction.payment_processor.void_transaction(transaction)
```
bogdanpetrea: I think you should check the result of this.

silver/models/payment_methods.py
```diff
@@ -84,6 +89,35 @@ def decrypt_data(self, crypted_data):
...
+            raise ValidationError("You can't cancel a canceled payment method.")
+
+        cancelable_states = [Transaction.States.Initial,
+                             Transaction.States.Pending]
+        transactions = Transaction.objects.filter(payment_method=self,
+                                                  state__in=cancelable_states)
+
+        for transaction in transactions:
+            if transaction.state == Transaction.States.Initial:
+                transaction.cancel()
```
bogdanpetrea: You should expect a `TransitionNotAllowed` exception here.

silver/models/payment_methods.py
```diff
@@ -84,6 +89,35 @@ def decrypt_data(self, crypted_data):
         except InvalidToken:
             return None
 
+    def cancel(self):
+        if self.canceled:
+            raise ValidationError("You can't cancel a canceled payment method.")
+
+        cancelable_states = [Transaction.States.Initial,
+                             Transaction.States.Pending]
+        transactions = Transaction.objects.filter(payment_method=self,
```
bogdanpetrea: You don't have to import Transaction for this. There probably is a related_name for the reverse relation on which can further filter.

silver/models/transactions/transaction.py
```diff
@@ -276,14 +277,16 @@ def _sync_transaction_state_with_document(transaction, target):
...
     payment_methods = PaymentMethod.objects.filter(
-        enabled=True,
+        canceled=False,
         verified=True,
         customer=document.customer
     )
     for payment_method in payment_methods:
         if (payment_method.verified and
-                payment_method.enabled):
+                not payment_method.canceled):
```
bogdanpetrea: I don't think this check is needed anymore because of the queryset above.

PR #501, opened 2017-02-01
Added methods for DocumentEntry currency conversion.

bogdanpetrea:


silver/models/documents/base.py
```diff
@@ -469,3 +469,47 @@ def serialize_hook(self, hook):
...
         }
+
+    @property
+    def entries(self):
+        raise NotImplementedError
+
+    @property
+    def total(self):
+        entries_total = [Decimal(entry.total) for entry in self.entries]
+        res = sum(entries_total)
```
vtemian: Just return sum

silver/models/documents/base.py
```diff
@@ -469,3 +469,47 @@ def serialize_hook(self, hook):
...
+        return res
+
+    @property
+    def total_before_tax(self):
+        entries_total = [Decimal(entry.total_before_tax)
+                         for entry in self.entries]
+        res = sum(entries_total)
```
vtemian: Same as above

PR #508, opened 2017-02-03
Added valid_until and display_info to PaymentMethod model.

vtemian:


silver/models/payment_methods.py
```diff
@@ -55,6 +55,9 @@ def as_list(cls):
     verified = models.BooleanField(default=False)
     canceled = models.BooleanField(default=False)
 
+    valid_until = models.DateTimeField(null=True, blank=True)
+    display_info = models.CharField(max_length=256, null=True, blank=True)
```
bogdanpetrea: What is this used for? `TextField` might be better.

vtemian: Just a small info to easily identify a payment method.
For PayPal should be the email and for credit card the last 4 digits.

vtemian: I really should exceed 256 characters.

bogdanpetrea: How about we use https://github.com/silverapp/silver/blob/master/silver/models/payment_methods.py#L140 which can get it's data from the `data` field. In this way we store data in one place and have it in a serializable form.

vtemian: I would pretty much keep it as a total separate think.

bogdanpetrea: Fair enough.

PR #519, opened 2017-02-07
Autocreate transactions on PaymentMethod verify

bogdanpetrea:


silver/api/serializers.py
```diff
@@ -458,6 +449,13 @@ def validate_payment_processor_name(self, value):
 
         return value
 
+    def validate_verified(self, value):
+        if self.instance and value is False and self.instance.verified is True:
```
vtemian: Remove actual boolean value and use python constructs.

silver/models/payment_methods.py
```diff
@@ -60,6 +62,11 @@ def as_list(cls):
 
     objects = InheritanceManager()
 
+    @property
+    def one_way_fields(self):
```
vtemian: Those are not tickets :))
Maybe something like `final_fields` or `locked_fields` or `imutable_fields`

silver/models/payment_methods.py
```diff
@@ -72,6 +79,10 @@ def __init__(self, *args, **kwargs):
             except AttributeError:
                 pass
 
+    @property
+    def transactions(self):
```
vtemian: Maybe is better to add a `related_name`

silver/models/payment_methods.py
```diff
@@ -126,14 +137,29 @@ def cancel(self):
 
         return None
 
-    def save(self, **kwargs):
-        self.clean()
-        super(PaymentMethod, self).save(**kwargs)
+    def clean_with_old_state(self, old_state):
```
vtemian: `previous_instance` ? In this project context, `state` can be easily confused with the field, not the object's state.

silver/models/payment_methods.py
bogdanpetrea: I know, but that's what they are. They're not final, otherwise you shouldn't be able to change them at all. And definitely not locked or imutable.

silver/models/payment_methods.py
bogdanpetrea: I just wanted to not have to call `.all()` everytime.

PR #521, opened 2017-02-07
Add action based view for transaction.

vtemian:


silver/api/serializers.py
```diff
@@ -496,7 +496,7 @@ class Meta:
                   'refund_code', 'cancel_code')
 
         read_only_fields = ('customer', 'provider', 'can_be_consumed', 'pay_url',
-                            'id', 'url', 'state', 'updated_at', 'created_at',
```
bogdanpetrea: Why not leave it as read_only?

vtemian: I will.
Those are some leftovers from the previous attempt.

silver/api/views.py
```diff
@@ -941,3 +941,35 @@ def get_object(self):
...
+
+class TransactionAction(APIView):
+    permission_classes = (permissions.IsAuthenticated,)
+
+    def post(self, request, *args, **kwargs):
+        transaction = self.get_object(**kwargs)
+        action_to_execute = getattr(transaction, kwargs.get('requested_action'),
+                                    None)
+
+        if not action_to_execute:
```
bogdanpetrea: Adding a list of allowed actions would be safer. The URL can still keep it's regex although it wouldn't be necessarily needed anymore. This is valid for the `PaymentMethodAction` view as well.

PR #523, opened 2017-02-08
Added pep8 linting requirements to CI build.

vtemian:


silver/tests/commands/test_generate_docs.py
```diff
@@ -661,7 +661,9 @@ def test_subscription_with_trial_with_metered_features_overflow_to_draft(self):
 
             doc = document_entries[4]
             assert doc.unit_price == metered_feature.price_per_unit
-            assert doc.quantity == units_consumed_during_trial - metered_feature.included_units_during_trial
+
+            trial_units = metered_feature.included_units_during_trial
```
bogdanpetrea: Maybe `included_trial_units` or `included_units` since `units_consumed_during_trial` are also `trial_units`.

silver/tests/spec/test_documents.py
```diff
@@ -41,9 +41,11 @@ def setUp(self):
 
     def _get_expected_data(self, document, transactions=None):
         kind = unicode(document.kind.lower())
+        payment_method_url = u'http://testserver/customers/%d/payment_methods/%d/'
```
bogdanpetrea: Doing `http://testserver + reverse(view, kwargs)` would be nice for all these urls, but it's not necessarily related to this PR.

silver/tests/unit/test_payment_method_serializer.py
```diff
@@ -41,12 +41,15 @@ def test_encoding(self):
             'request': request
         })
 
+        self_url_skelet = 'http://testserver/customers/{}/payment_methods/{}/'
```
bogdanpetrea: Maybe something like this would help : https://github.com/silverapp/silver/blob/master/silver/tests/spec/test_transactions.py#L40
`skelet` sounds a bit odd.

PR #532, opened 2017-05-19
Pdf generation task

bogdanpetrea:
* Added a `PDF` model
* Document `pdf` field is no longer a `FileField` but a `ForeignKey` to the `PDF` model
* Document's are automatically saved after every transition for more consistency
* PDF's are now generated and uploaded asynchronously through a non-overlapping celery beat task
* Made sure the migrations apply properly (forwards and backwards)

### DIFF
--- a/silver/migrations/0036_auto_20170514_1627.py
+++ b/silver/migrations/0036_auto_20170514_1627.py
@@ -2,6 +2,7 @@
 from __future__ import unicode_literals
 
 import uuid
+from itertools import chain
 
 from django.db import migrations, models
 
@@ -22,20 +23,29 @@ class Migration(migrations.Migration):
         Proforma = apps.get_model('silver', 'Proforma')
         PDF = apps.get_model('silver', 'PDF')
 
-        for invoice in Invoice.objects.using(db_alias).exclude(
-            state=BillingDocumentBase.STATES.DRAFT
+        for document in chain(
+            Invoice.objects.using(db_alias).exclude(state=BillingDocumentBase.STATES.DRAFT),
+            Proforma.objects.using(db_alias).exclude(state=BillingDocumentBase.STATES.DRAFT),
         ):
-            pdf_object = PDF.objects.using(db_alias).create(
-                upload_path=invoice.get_pdf_upload_path()
-            )
-            pdf_object.pdf_file = invoice.pdf
+            pdf_object = PDF.objects.using(db_alias).create()
+            pdf_object.pdf_file = document.pdf_old
             pdf_object.save(using=db_alias)
 
-            invoice.pdf = pdf_object
-            invoice.save(using=db_alias)
+            document.pdf = pdf_object
+            document.save(using=db_alias)
 
     def move_pdf_from_model_to_documents(apps, schema_editor):
-        pass
+        db_alias = schema_editor.connection.alias
+
+        Invoice = apps.get_model('silver', 'Invoice')
+        Proforma = apps.get_model('silver', 'Proforma')
+
+        for document in chain(
+            Invoice.objects.using(db_alias).exclude(state=BillingDocumentBase.STATES.DRAFT),
+            Proforma.objects.using(db_alias).exclude(state=BillingDocumentBase.STATES.DRAFT),
+        ):
+            document.pdf_old = document.pdf.pdf_file
+            document.save()
 
     operations = [
         migrations.CreateModel(
@@ -81,7 +91,9 @@ class Migration(migrations.Migration):
             model_name='proforma',
             name='pdf_old',
         ),
-        migrations.RunSQL("""
+
+        migrations.RunSQL(
+            sql="""
                 DROP VIEW IF EXISTS silver_document;
                 CREATE VIEW silver_document AS SELECT
                     'invoice' AS `kind`, id, series, number, issue_date, due_date,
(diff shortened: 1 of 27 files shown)


silver/migrations/0036_auto_20170514_1627.py
```diff
@@ -0,0 +1,100 @@
...
+    ]
+
+    def move_pdf_from_documents_to_model(apps, schema_editor):
+        db_alias = schema_editor.connection.alias
+
+        Invoice = apps.get_model('silver', 'Invoice')
+        Proforma = apps.get_model('silver', 'Proforma')
+        PDF = apps.get_model('silver', 'PDF')
+
+        for invoice in Invoice.objects.using(db_alias).filter(
```
vtemian: I find `exclude` easier to read.

silver/migrations/0036_auto_20170514_1627.py
```diff
@@ -0,0 +1,100 @@
...
+            ~Q(state=BillingDocumentBase.STATES.DRAFT)
+        ):
+            pdf_object = PDF.objects.using(db_alias).create(
+                upload_path=invoice.get_pdf_upload_path()
```
vtemian: Why don't you create the pdf with the `pdf_file` already set up?
Do you need to call `save()` ?

silver/migrations/0036_auto_20170514_1627.py
```diff
@@ -0,0 +1,100 @@
...
+            invoice.save(using=db_alias)
+
+    def move_pdf_from_model_to_documents(apps, schema_editor):
+        pass
+
+    operations = [
+        migrations.CreateModel(
+            name='PDF',
+            fields=[
+                ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
```
vtemian: I would add an `UUID` field as well.

silver/models/documents/base.py
```diff
@@ -419,38 +422,63 @@ def get_template(self, state=None):
...
+            'doc': self,
+            'filename': self.get_pdf_filename(),
+            'provider': self.provider.slug,
+            'customer': self.customer.slug,
+            'issue_date': self.issue_date.strftime('%Y/%m/%d')
+        }
+
+        return path_template.format(**context)
+
+    def generate_pdf(self, state=None, upload=True):
```
vtemian: I would make `PDF` a `OneToOne` field and move the entire pdf generation logic to pdf model.
`get_pdf_filename`, `get_pdf_upload_path` and `generate_pdf`

silver/models/documents/base.py
```diff
@@ -493,3 +521,72 @@ def total_before_tax_in_transaction_currency(self):
...
+
+    document = instance
+    setattr(document, '.recently_transitioned', target)
+
+    document.save()
+
+
+@receiver(post_save)
+def post_document_save(sender, instance, created=False, **kwargs):
+    if not isinstance(instance, BillingDocumentBase):
```
vtemian: `if not isinstance(instance, BillingDocumentBase) or not hasattr(instance, '.recently_transitioned'):`
This will improve readability a little bit.

silver/models/documents/pdf.py
```diff
@@ -0,0 +1,55 @@
...
+def get_storage():
+    storage_settings = getattr(settings, 'SILVER_DOCUMENT_STORAGE', None)
+    if not storage_settings:
+        return
+
+    storage_class = import_string(storage_settings[0])
+    return storage_class(*storage_settings[1], **storage_settings[2])
+
+
+def get_upload_path(instance, filename):
```
vtemian: Do you still need this one?

silver/models/documents/pdf.py
```diff
@@ -0,0 +1,55 @@
...
+
+class PDF(Model):
+    pdf_file = FileField(null=True, blank=True, editable=False,
+                         storage=get_storage(), upload_to=get_upload_path)
+    dirty = BooleanField(default=False)
+    upload_path = TextField(null=True, blank=True)
+
+    @property
+    def url(self):
+        return self.pdf_file.url if self.pdf_file else None
```
vtemian: I would return an empty string to be consistent with the returning type.

silver/tasks.py
```diff
@@ -0,0 +1,38 @@
...
+        document = Invoice.objects.get(id=document_id)
+    else:
+        document = Proforma.objects.get(id=document_id)
+
+    document.generate_pdf()
+
+
+@task()
+def generate_pdfs():
+    lock = lock_manager.lock('generate_pdfs', 60000)
```
vtemian: Magic number in here

silver/tasks.py
```diff
@@ -0,0 +1,38 @@
...
+    if not lock:
```
vtemian: I would retry it somehow.

silver/tasks.py
```diff
@@ -0,0 +1,38 @@
...
+        return
+
+    tasks = []
+
+    dirty_documents = chain(Invoice.objects.filter(pdf__dirty=True),
+                            Proforma.objects.filter(pdf__dirty=True))
+
+    async_results = list([
+        tasks.append(generate_pdf.delay(document.id, document.kind))
```
vtemian: `append` returns `None` and this will look like a list of Nones

vtemian: btw, you can use a `group` http://docs.celeryproject.org/en/latest/userguide/canvas.html#groups

silver/tests/integration/test_documents_transactions.py
```diff
@@ -106,17 +106,14 @@ def test_no_transaction_creation_for_issued_documents_case3(self):
             document.
         """
         invoice = InvoiceFactory.create()
-        invoice.issue()
+        invoice.issue()  # this creates a Transaction
```
vtemian: :+1:

silver/models/documents/pdf.py
bogdanpetrea: Yes, it is used below in `pdf_file`.

silver/models/documents/pdf.py
bogdanpetrea: When would we ever return `None` then? The empty string could be considered a valid relative url.

silver/tasks.py
bogdanpetrea: It is retried by celery beat.

silver/models/documents/base.py
bogdanpetrea: I think it's ok to pass filename and upload_path to the PDF model when creating an object.
And generate_pdf contains some BillingDocuments related logic.

PR #533, opened 2017-05-22
Split payments

bogdanpetrea:
Depends on #532.

### DIFF
--- a/silver/models/documents/base.py
+++ b/silver/models/documents/base.py
@@ -28,7 +28,8 @@ from django.conf import settings
 from django.core.exceptions import ValidationError, NON_FIELD_ERRORS
 from django.core.urlresolvers import reverse
 from django.core.validators import MinValueValidator
-from django.db import models, transaction
+from django.db import models
+from django.db import transaction as db_transaction
 from django.db.models import Max, ForeignKey
 from django.template.loader import select_template
 from django.utils import timezone
@@ -305,7 +306,7 @@ class BillingDocumentBase(models.Model):
 
         self._last_state = self.state
 
-        with transaction.atomic():
+        with db_transaction.atomic():
             # Create pdf object
             if not self.pdf and self.state != self.STATES.DRAFT:
                 self.pdf = PDF.objects.create(upload_path=self.get_pdf_upload_path(), dirty=True)
@@ -462,7 +463,7 @@ class BillingDocumentBase(models.Model):
         pdf_file_object = self.pdf.generate(template=self.get_template(state),
                                             context=context,
                                             upload=upload)
-        with transaction.atomic():
+        with db_transaction.atomic():
             # lock pdf
             PDF.objects.select_for_update().filter(pk=self.pdf.pk)
 
@@ -526,6 +527,33 @@ class BillingDocumentBase(models.Model):
     def tax_value_in_transaction_currency(self):
         return sum([entry.tax_value_in_transaction_currency
                     for entry in self.entries])
+
+    @property
+    def amount_paid_in_transaction_currency(self):
+        Transaction = get_model('silver.Transaction')
+
+        return sum([transaction.amount
+                    for transaction in self.transactions.filter(state=Transaction.States.Settled)])
+
+    @property
+    def amount_pending_in_transaction_currency(self):
+        Transaction = get_model('silver.Transaction')
+
+        return sum([transaction.amount
+                    for transaction in self.transactions.filter(state=Transaction.States.Pending)])
+
+    @property
+    def amount_to_be_charged_in_transaction_currency(self):
+        Transaction = get_model('silver.Transaction')
+
+        return self.total_in_transaction_currency - sum([
+            transaction.amount
+            for transaction in self.transactions.filter(state__in=[
+                Transaction.States.Initial,
+                Transaction.States.Pending,
+                Transaction.States.Settled
+            ])
+        ])
 
 
 def create_transaction_for_document(document):
--- a/silver/models/documents/base.py
+++ b/silver/models/documents/base.py
@@ -281,6 +281,11 @@ class BillingDocumentBase(models.Model):
             msg = 'You cannot edit the document once it is in paid state.'
             raise ValidationError({NON_FIELD_ERRORS: msg})
 
+        if self.transactions.exclude(currency=self.transaction_currency).exists():
+            message = 'There are unfinished transactions of this document that use a ' \
+                      'different currency.'
+            raise ValidationError({'transaction_currency': message})
+
     def save(self, *args, **kwargs):
         if not self.transaction_currency:
             self.transaction_currency = self.customer.currency or self.currency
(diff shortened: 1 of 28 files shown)


silver/models/documents/base.py
```diff
@@ -493,3 +527,99 @@ def total_before_tax_in_transaction_currency(self):
...
+        return
+
+    document = instance
+
+    if hasattr(document, '.recently_transitioned'):
+        delattr(document, '.recently_transitioned')
+
+        # Transition related document too, if needed
+        if document.related_document and document.state != document.related_document.state:
+            state_transition_map = {
```
vtemian: Seems like a useful business logic block that can be extracted in `BillingDocumentBase` model

silver/models/documents/base.py
```diff
@@ -493,3 +527,99 @@ def total_before_tax_in_transaction_currency(self):
...
+                BillingDocumentBase.STATES.ISSUED: 'issue',
+                BillingDocumentBase.STATES.CANCELED: 'cancel',
+                BillingDocumentBase.STATES.PAID: 'pay'
+            }
+            transition_name = state_transition_map[document.state]
+
+            bound_transition_method = getattr(document.related_document, transition_name)
+            bound_transition_method()
```
vtemian: This will also save the `document.related_document` instance?

vtemian: Maybe we can think to move this logic into a reconcile loop. (not necessarily now)

silver/models/documents/base.py
```diff
@@ -493,3 +527,99 @@ def total_before_tax_in_transaction_currency(self):
...
+
+
+@receiver(post_save)
+def post_document_save(sender, instance, created=False, **kwargs):
+    if not isinstance(instance, BillingDocumentBase):
+        return
+
+    document = instance
+
+    if hasattr(document, '.recently_transitioned'):
```
vtemian: Negating this will make the entire logic a little easier to read.

silver/models/documents/base.py
```diff
@@ -493,3 +527,99 @@ def total_before_tax_in_transaction_currency(self):
...
+            # The related document might have the only reference to an existing transaction
+            if not (document.related_document or document).transactions.filter(
+                state__in=[Transaction.States.Pending,
+                           Transaction.States.Initial,
+                           Transaction.States.Settled]
+            ):
+                create_transaction_for_document(document)
+
+        # Generate a PDF
+        document.set_pdf_for_generation()
```
vtemian: :+1:
I would name it `mark_for_generation` (in this way you will leave some options to have more renders: pdf, html, xml. Maybe, in a near feature, each invoice will have a well know format, xml based, making them easier to import/export (I think that the Czech Republic has this system implemented)).

silver/models/documents/pdf.py
```diff
@@ -0,0 +1,58 @@
...
+                         storage=get_storage(), upload_to=get_upload_path)
+    dirty = BooleanField(default=False)
+    upload_path = TextField(null=True, blank=True)
+
+    @property
+    def url(self):
+        return self.pdf_file.url if self.pdf_file else None
+
+    def generate(self, template, context, upload=True):
+        pdf_file_object = HttpResponse(content_type='application/pdf')
```
vtemian: Why do you need to return an `HttpResponse` from model.
I don't think that models should be coupled with any type of http response.

silver/models/transactions/transaction.py
```diff
@@ -136,6 +136,43 @@ def refund(self, refund_code='default', refund_reason='Unknown refund reason'):
...
+                lock_key = 'proforma_{pk}'.format(pk=self.proforma.pk)
+            elif self.invoice:
+                lock_key = 'invoice_{pk}'.format(pk=self.invoice.pk)
+            else:
+                raise ValidationError(
+                    'The transaction must have at least one document '
+                    '(invoice or proforma).'
+                )
+
+            lock = redis.lock(lock_key, timeout=5)  # 5 seconds should be more than enough to save
```
vtemian: Magic number detected

silver/models/transactions/transaction.py
```diff
@@ -136,6 +136,43 @@ def refund(self, refund_code='default', refund_reason='Unknown refund reason'):
         self.refund_code = refund_code
         logger.error(str(refund_reason))
 
+    def save(self, *args, **kwargs):
+        previous_instance = get_object_or_None(Transaction, pk=self.pk) if self.pk else None
+        setattr(self, 'previous_instance', previous_instance)
+
+        def _clean_and_save():
+            if not getattr(self, '.cleaned', False):
```
vtemian: It looks like a method.
Do you use the closure, or you just want to reuse the code?

silver/models/transactions/transaction.py
```diff
@@ -136,6 +136,43 @@ def refund(self, refund_code='default', refund_reason='Unknown refund reason'):
...
+            else:
+                raise ValidationError(
+                    'The transaction must have at least one document '
+                    '(invoice or proforma).'
+                )
+
+            lock = redis.lock(lock_key, timeout=5)  # 5 seconds should be more than enough to save
+
+            if not lock.acquire(blocking_timeout=5):
+                raise ValidationError("Couldn't create a transaction for the given document.")
```
vtemian: It doesn't look like a `ValidationError`.
Also, maybe some details regarding why the transaction can't be created will help debugging.

silver/models/transactions/transaction.py
```diff
@@ -136,6 +136,43 @@ def refund(self, refund_code='default', refund_reason='Unknown refund reason'):
...
+                lock_key = 'invoice_{pk}'.format(pk=self.invoice.pk)
+            else:
+                raise ValidationError(
+                    'The transaction must have at least one document '
+                    '(invoice or proforma).'
+                )
+
+            lock = redis.lock(lock_key, timeout=5)  # 5 seconds should be more than enough to save
+
+            if not lock.acquire(blocking_timeout=5):
```
vtemian: Magic number detected.

silver/models/documents/base.py
bogdanpetrea: It could be...

silver/models/documents/base.py
bogdanpetrea: 1. Yes, it will also save the related document, because the `save` method gets automatically called after a successful transition.
2. I'd rather have this happen synchronously. Having a draft proforma and an issued invoice, even for a short period of time, doesn't sound good.

silver/models/transactions/transaction.py
bogdanpetrea: I just want to reuse the code, but I don't want to expose it, because calling that method would evict the locking.

silver/models/documents/pdf.py
bogdanpetrea: That HttpResponse will be used as a PDF file_object for the `generate_pdf_template_object` method.

silver/models/transactions/transaction.py
bogdanpetrea: Well it's a validation step that doesn't pass. How about: `raise ValidationError("A transaction for the same billing document is currently being created.")`?

PR #538, opened 2017-07-14
Retry transactions

bogdanpetrea:


auvipy:
can we expect this to resume?

bogdanpetrea:
Hey @auvipy,

Except for bug or security fixes, Silver is not being actively developed at the moment.
Contributions in the form of ideas and pull requests will be handled, but not in priority mode.

Currently, I am the sole maintainer of this project, meaning unless somebody is going to pick up this feature, it is probably not going to be implemented anytime soon.

The requirements/specifications for this feature are also lacking, so contributions in that area would also be required.

If anybody is willing to tackle this problem, I am going to assist along the way.

auvipy:
OK, I will look into it to take over.

PR #539, opened 2017-07-18
Seed command (closes #140)

### LINKED ISSUE #140: Generate initial data
Generate some initial data for testing purposes using the predefined factories. The generated data should be loaded automatically after `syncdb`. 

Either make a script which uses the factories and is run after `syncdb` or using the factories generate some data and add it to `initial_data.json`.

vasiioandre:


silver/management/commands/seed.py
```diff
@@ -0,0 +1,54 @@
...
+
+from silver.models import Transaction, Invoice, Proforma
+from silver.tests.factories import (TransactionFactory, InvoiceFactory, DocumentEntryFactory,
+                                    ProformaFactory)
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = 'Adding paid invoices.'
```
bogdanpetrea: Use a better description for this command.

silver/tests/spec/test_proforma.py
```diff
@@ -139,7 +139,7 @@ def test_get_proforma(self):
         self.assertEqual(response.data, {
             "id": proforma.pk,
             "series": "ProformaSeries",
-            "number": 1,
+            "number": None,
```
bogdanpetrea: Maybe use `proforma.number`.

PR #544, opened 2017-08-11
WIP: Stats

vasiioandre:


stats/api/views.py
```diff
@@ -8,6 +12,28 @@
 from datetime import datetime, timedelta
 
 
+class HomeView(View):
```
bogdanpetrea: This class name isn't very suggestive. Maybe `StatsView` / `ChartsView` would be better suited.

stats/stats.py
```diff
@@ -74,23 +74,29 @@ class Stats(object):
             }
         },
         'subscriptions': {
+            'estimated_income':{
                 'modifiers': ['include_unused_plans'],
                 'granulations': {
-                    'plan': None,
-                    'month': ['Jan']
+                    'billing_date': ['year', 'month'],
```
bogdanpetrea: `year` and `month` are periods or intervals. Maybe rename `billing_date` to `billing_period` or something simillar.

stats/stats.py
```diff
@@ -74,23 +74,29 @@ class Stats(object):
...
+            raise ValueError('invalid model : choices are: Transaction, Invoice, Subscription')
         if self.queryset.model == Invoice:
             model = 'documents'
         elif self.queryset.model == Transaction:
             model = 'transactions'
+        else:
+            model = 'subscriptions'
         if self.result_type is None:
-            raise ValueError('Select a result type!')
+            raise ValueError('Select a result type: choices are: amount, count, overdue_payments, '
```
bogdanpetrea: You should fetch these result types dinamically.

stats/stats.py
```diff
@@ -222,40 +227,46 @@ def documents_overdue_payments(self, queryset):
...
+        }
+        stats_list.append(data)
+        return stats_list
+
     def subscriptions_estimated_income(self, queryset):
         entries = BillingLog.objects.filter(subscription_id=OuterRef('pk')).order_by().\
-        values('subscription_id').annotate(sum=Sum('total')).values('sum')
+            values('subscription_id').annotate(sum=Sum('total')).values('sum')
 
         final = queryset.annotate(
```
bogdanpetrea: Find a better name for this queryset. `final` is not that representative.

stats/stats.py
```diff
@@ -222,40 +227,46 @@ def documents_overdue_payments(self, queryset):
...
-            ), Decimal('0.00')) + F('plan__amount')
+            ), Decimal('0.00'))
         ).values(
             'subscription_amount', 'id', 'plan', 'plan__name', 'customer__first_name',
             'customer__last_name'
         ).order_by('-subscription_amount')
 
-        tupleList = []
+        stats_list = []
         for i in final:
```
bogdanpetrea: Find a better name for each element in the queryset, instead of just `i`.

stats/stats.py
```diff
@@ -222,40 +227,46 @@ def documents_overdue_payments(self, queryset):
...
-                }
-                tupleList.append(data)
+            if not stats_list:
+                stats_list = self.add_list_item(stats_list, i['plan__name'], i['plan'],
+                                                i['id'], i['subscription_amount'],
+                                                i['customer__first_name'] + " " + i['customer__last_name'])
             else:
                 ok = 0
-                for item in tupleList:
+                for item in stats_list:
```
bogdanpetrea: Find a better name for the elements in stats_list. `item` isn't suggestive enough.

stats/stats.py
```diff
@@ -222,40 +227,46 @@ def documents_overdue_payments(self, queryset):
...
                     for key, value in item.iteritems():
```
bogdanpetrea: Try to find better names instead of `key`, `value`. What do these represent?

stats/templates/base.html
```diff
@@ -0,0 +1,27 @@
...
+        <script>
+            $(document).ready(function(data){
+                {% block jquery %}
+                {% endblock %}
+            })
+        </script>
+    </div>
+
+    </body>
+</html>
```
bogdanpetrea: You are missing a newline here.

stats/templates/base/bootstrap.html
```diff
@@ -0,0 +1,4 @@
+ <!--[if lt IE 9]>
+  <script src="https://oss.maxcdn.com/html5shiv/3.7.3/html5shiv.min.js"></script>
+  <script src="https://oss.maxcdn.com/respond/1.4.2/respond.min.js"></script>
+<![endif]-->
```
bogdanpetrea: You are missing a newline here.

stats/templates/charts.html
```diff
@@ -0,0 +1,104 @@
...
+        <div class="chartWrapper">
+          <div class="chartAreaWrapper">
+            <div class="chartAreaWrapper2">
+                <canvas id="canvas"></canvas>
+            </div>
+          </div>
+
+            <canvas id="myChartAxis" height="200" width="0"></canvas>
+        </div>
+    {% endblock content %}
```
bogdanpetrea: You are missing a newline here.

stats/api/views.py
```diff
@@ -0,0 +1,77 @@
...
+from stats.stats import Stats
+from datetime import datetime, timedelta
+
+
+class HomeView(View):
+    def get(self, request, *args, **argv):
+        return render(request, 'charts.html', {})
+
+
+class SubscriptionStatsChart(APIView):
```
bogdanpetrea: What is this going to be used for?

stats/api/views.py
```diff
@@ -0,0 +1,77 @@
...
+    authentication_classes = []
+    permission_classes = []
+
+    def get(self, request, format=None):
+        labels = ["plan1", "plan2", "plan3", "plan4", "plan5"]
+        default_items = [1, 6, 9, 3, 1]
+        default_items2 = [2, 2, 3, 9]
+        data = {
```
bogdanpetrea: This looks like it's been written for a specific kind of client consumer, which is not a good idea. But first I'll wait for you to reveal the purpose of this class.

stats/stats.py
```diff
@@ -0,0 +1,324 @@
...
+from silver.models import Transaction, Invoice, Proforma, MeteredFeatureUnitsLog, Subscription, \
+    Plan, BillingLog
+from datetime import datetime, date, timedelta
+from django.db.models.functions import ExtractYear, ExtractMonth, ExtractWeekDay, ExtractWeek, \
+    Coalesce
+from django.db.models import Avg, Sum, Count, DecimalField, F, Case, When, Max, Subquery, OuterRef
+from django.db.models import ExpressionWrapper
+
+{
+    'granulations': [
```
bogdanpetrea: I think this can be safely removed.

stats/stats.py
```diff
@@ -0,0 +1,324 @@
+from decimal import Decimal
+
+from silver.models import Transaction, Invoice, Proforma, MeteredFeatureUnitsLog, Subscription, \
+    Plan, BillingLog
+from datetime import datetime, date, timedelta
+from django.db.models.functions import ExtractYear, ExtractMonth, ExtractWeekDay, ExtractWeek, \
```
bogdanpetrea: Please group imports that span over more than one line, within parantheses.
```
from django.db.models.functions import (ExtractYear, ExtractMonth, ExtractWeekDay, ExtractWeek,
                                                                 Coalesce)
```

stats/stats.py
```diff
@@ -0,0 +1,324 @@
...
+
+        for key, value in self.result_types.iteritems():
+            if key != model:
+                continue
+            if self.result_type not in value:
+                raise ValueError('Invalid result type!')
+            for result_type, values in value.iteritems():
+                if result_type == self.result_type:
+                    if self.granulations['value'] not in values:
+                        raise ValueError('Invalid granulation!')
```
bogdanpetrea: Please don't use exclamation marks.

stats/stats.py
```diff
@@ -0,0 +1,324 @@
...
+                    elif not values['modifiers']:
+                        if self.modifier is not None:
+                            raise ValueError("This result type doesn't have modifiers!")
+                    elif self.modifier is None:
+                            raise ValueError('Select a modifier!')
```
bogdanpetrea: As a general advice, exception messages should be impersonal. Something like this would be better:
```
raise ValueError('The modifier argument is required.')
```

stats/stats.py
```diff
@@ -0,0 +1,324 @@
...
+                continue
+            if self.result_type not in value:
+                raise ValueError('Invalid result type!')
+            for result_type, values in value.iteritems():
+                if result_type == self.result_type:
+                    if self.granulations['value'] not in values:
+                        raise ValueError('Invalid granulation!')
+                    elif not values['modifiers']:
+                        if self.modifier is not None:
+                            raise ValueError("This result type doesn't have modifiers!")
```
bogdanpetrea: `The` might be better than `This`.

stats/stats.py
```diff
@@ -0,0 +1,324 @@
...
+
+    def documents_overdue_payments(self, queryset):
+        overdue = {}
+        for i in queryset.filter(due_date__lte=datetime.now()):
+            new = {i.customer: [{'total': i.total, 'paid': i.amount_paid_in_transaction_currency,
+                                 'left': i.amount_to_be_charged_in_transaction_currency}]}
+            overdue[i.customer_id] = new
+        return overdue
+
+    def add_list_item(self, stats_list, plan_name, plan_id, subscription_id, subscription_amount, customer):
```
bogdanpetrea: This is only called within the `subscriptions_estimated_income` method. You should rename it and maybe precede it with an underscore. Something like `_add_estimated_income_to_list`.

stats/stats.py
```diff
@@ -0,0 +1,324 @@
...
+            'values': [
+                {
+                    'subscription_id': subscription_id,
+                    'estimated_income': subscription_amount,
+                    'customer_name': customer
+                }
+            ]
+        }
+        stats_list.append(data)
+        return stats_list
```
bogdanpetrea: You don't have to return the list, as lists in Python are mutable and variables only have the references to the actual list.

stats/stats.py
```diff
@@ -0,0 +1,324 @@
...
+                                                i['id'], i['subscription_amount'],
+                                                i['customer__first_name'] + " " + i['customer__last_name'])
+            else:
+                ok = 0
+                for item in stats_list:
+                    for key, value in item.iteritems():
+                        if key == 'granulations':
+                            for keys, values in value.iteritems():
+                                if values['name'] == i['plan__name']:
+                                    item['values'].append(
```
bogdanpetrea: Why are you not using the helper method you wrote above for this?

stats/stats.py
```diff
@@ -0,0 +1,324 @@
...
+                                        }
+                                    )
+                                    ok = 1
+                if ok == 0:
+                    stats_list = self.add_list_item(stats_list, i['plan__name'], i['plan'], i['id'],
+                                                    i['subscription_amount'],
+                                                    i['customer__first_name'] + " " + i[
+                                                        'customer__last_name'])
+
+        if self.modifier is not None:
```
bogdanpetrea: You should look for the specific `include_unused_plans` here, not just any modifier.

stats/stats.py
```diff
@@ -0,0 +1,324 @@
...
+                        'id': i.id
+                    }},
+                    'values': []
+                }
+                stats_list.append(data)
+
+        return stats_list
+
+    def get_result(self):
+        if self.queryset.model == Transaction:
```
bogdanpetrea: You should build these dynamically and avoid all the if branches. Have a method that receives a model and returns the model name (e.g. `Transaction` -> `transactions`) and use that to obtain your method name.

stats/api/views.py
vasiioandre: It's used to render the chart from the template charts.html

stats/stats.py
vasiioandre: I don't understand what you mean, how can I fetch them dinamically?

stats/stats.py
vasiioandre: Because I want to add only a subscription to an existing plan in the list, whereas in that helper method  I add a new plan and his first subscription

stats/api/views.py
```diff
@@ -34,18 +34,16 @@ def list(self, request, *args, **kwargs):
...
 
-        if query_params.get('granulation_plan', None) is not None:
-            granulations_list.append({'name': 'plan',
-                                      'value': None})
-        if query_params.get('granulation_customer', None) is not None:
-            granulations_list.append({'name': 'customer',
-                                      'value': None})
+        for url_parameter, url_parameter_values in request.GET.lists():
+            if url_parameter != 'result_type' and url_parameter != 'modifier':
+                granulations_list.append({'name': url_parameter[12:],
```
bogdanpetrea: You shouldn't rely on magic numbers.

stats/stats.py
```diff
@@ -64,122 +64,112 @@ def __init__(self, queryset, result_type, modifier=None, granulations=None):
...
         if self.queryset.model != Transaction and self.queryset.model != Invoice \
                 and self.queryset.model != Subscription and self.queryset.model != BillingLog:
-            raise ValueError('Invalid model, choices are: Transaction, Invoice, Subscription, '
-                             'BillingLog')
+            raise ValueError('Invalid model, choices are: Transaction, Invoice, BillingLog')
         if self.result_type is None:
-            raise ValueError('Select a result type: choices are: amount, count, overdue_payments, '
-                             'estimated_income')
+            raise ValueError('Select a result type: choices are: amount, overdue_payments, '
+                             'paid_date, total')
```
bogdanpetrea: The valid result types shouldn't be hardcoded.

stats/stats.py
```diff
@@ -64,122 +64,112 @@ def __init__(self, queryset, result_type, modifier=None, granulations=None):
...
+                if granulation_parameters and granulation['value'] not in granulation_parameters:
+                    raise ValueError("The granulation parameter is incorrect: "
+                                     "Try granulation_field=month")
+
     def validate(self):
         if self.queryset.model != Transaction and self.queryset.model != Invoice \
                 and self.queryset.model != Subscription and self.queryset.model != BillingLog:
-            raise ValueError('Invalid model, choices are: Transaction, Invoice, Subscription, '
-                             'BillingLog')
+            raise ValueError('Invalid model, choices are: Transaction, Invoice, BillingLog')
```
bogdanpetrea: The valid models shouldn't be hardcoded.

stats/stats.py
```diff
@@ -64,122 +64,112 @@ def __init__(self, queryset, result_type, modifier=None, granulations=None):
...
+        search_tuple_key = granulated_stats.get(tuple_key, [])
 
         details = {'id': id}
         if amount is not None:
             details['total'] = amount
         if 'additional_granulation_field' not in granulation_arguments:
             details['customer_name'] = customer_name
         if additional_field is not None:
-            details[next(iter(additional_field))] = additional_field[next(iter(additional_field))]
+            details[additional_field.items()[0][0]] = additional_field.items()[0][1]
```
bogdanpetrea: Please leave a comment here on what this line does.

stats/stats.py
```diff
@@ -191,41 +181,57 @@ def _name_tuple(self, granulation_arguments):
...
+
         if 'granulation_field' in granulation_arguments:
             if 'additional_granulation_field' in granulation_arguments:
-                item = name_tuple(granulation=granulation_value, name=customer_name,
-                                  currency=currency)
-                self._add_item_to_list(item, doc, granulation_arguments, customer_name, amount, id,
-                                       additional_field)
+                tuple_key = name_tuple(granulation=granulation_value, name=customer_name,
+                                       currency=currency)
+                self._search_element(tuple_key, granulated_stats, granulation_arguments,
```
bogdanpetrea: I don't find `_search_element` to be descriptive enough.

stats/stats.py
```diff
@@ -191,41 +181,57 @@ def _name_tuple(self, granulation_arguments):
...
-            data['currency'] = key.currency
-            if 'additional_granulation_field' in granulation_arguments:
-                data['customer_name'] = key.name
-            if 'granulation_field' in granulation_arguments:
-                data[granulation_field] = key.granulation
-            data['values'] = []
-            for subscription in sorted(doc[key]):
-                data['values'].append(subscription)
-            stats_list.append(data)
+    # returns the formatted granulation date according to the selected granulation parameter
```
bogdanpetrea: Move this commentary inside the method as a docstring.

stats/stats.py
```diff
@@ -271,16 +278,20 @@ def documents_paid_date(self, queryset):
             else:
                 granulation_value = None
 
-            self.get_stats_data(granulation_arguments, customer_name, document.currency, None,
-                                granulation_value, document.id, doc, None)
+            additional_field = {"payment_day": document.paid_date.strftime('%d')}
+
+            self.get_granulation_tuple(granulation_arguments, customer_name, document.currency,
```
bogdanpetrea: Because there are many arguments in this call, it would be nice to specify their name, in other words send them as keyword args.

stats/stats.py
```diff
@@ -0,0 +1,403 @@
...
+                return namedtuple("group_by", ["name", "granulation", "currency"])
+            else:
+                return namedtuple("group_by", ["granulation", "currency"])
+        elif 'additional_granulation_field' in granulation_arguments:
+            return namedtuple("group_by", ["name", "currency"])
+        else:
+            return namedtuple("group_by", ["currency"])
+
+    # gets a tuple formed by the selected granulations
+    def get_granulation_tuple(self, granulation_arguments, customer_name, currency, amount,
```
bogdanpetrea: Try not to pass each field that's contained in the stats separately, such as `amount`, `currency`, `customer`... Maybe put them in a dict and pass them like that.

stats/stats.py
```diff
@@ -64,122 +64,112 @@ def __init__(self, queryset, result_type, modifier=None, granulations=None):
...
-            7: "Jul",
-            8: "Aug",
-            9: "Sep",
-            10: "Oct",
-            11: "Nov",
-            12: "Dec",
-        }
-        return switcher.get(argument, "none")
+    # iterates the granulated_stats dictionary and creates the stats_list
+    def _create_stats_list(self, granulated_stats, stats_list, granulation_field,
```
bogdanpetrea: You don't need to receive `stats_list`, you can just return a new list from this method.

stats/stats.py
```diff
@@ -64,122 +64,112 @@ def __init__(self, queryset, result_type, modifier=None, granulations=None):
...
-                month = self._numbers_to_strings(getattr(document, granulation_field).month)
-                return str(getattr(document, granulation_field).year) + ' ' + month
-            elif granulation_arguments['time_granulation_interval'] == 'day':
-                day = self._numbers_to_strings(
-                    getattr(document, granulation_field).month) + ' ' + str(
-                    getattr(document, granulation_field).day)
-                return str(getattr(document, granulation_field).year) + ' ' + day
+    # search the elements that correspond with the granulation tuple key given
+    def _search_element(self, tuple_key, granulated_stats, granulation_arguments, customer_name,
+                        amount, id, additional_field):
```
bogdanpetrea: You don't need to receive the `granulated_stats` list here. Just return the element and append it to the granulated_stats one level up in the call tree.

stats/stats.py
```diff
@@ -361,11 +389,15 @@ def _get_model_name(self):
             return 'billing_log'
 
     def get_result(self):
-        if self.modifier is not None:
-            method_name = self._get_model_name() + '_' + self.result_type + '_' + self.modifier
-            method = getattr(self, method_name)
-            return method(self.queryset)
+        isValid = self.validate()
```
bogdanpetrea: In python the convention is to use snake case for naming variables: `is_valid`

stats/tests/api/test_stats.py
```diff
@@ -47,27 +30,21 @@ def test_stats_billing_log_view_is_correct(api_client, create_subscription):
...
-    lista = [{'name': 'plan', 'value': None}, {'name': 'customer', 'value': None}]
-    stats = Stats(BillingLog.objects.all(), 'total', None, lista)
-    for i in BillingLog.objects.all():
-        print i.subscription.customer.first_name, i.subscription.plan
-    print stats.validate()
-
     assert response.data == [
-        {'currency': u'USD', 'values': [{'total': Decimal('20.00'),
-                                         'billing_date': datetime.date(2017, 1, 31), 'id': 2}],
+        {'currency': u'USD', 'values': [
```
bogdanpetrea: Try to keep one key: value pair per row inside dicts:
```
{
    'currency': u'USD',
    'values': [{
        ...
    }]
}
```

stats/fixtures/pytest_fixtures.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+@pytest.fixture()
+def api_client(user):
+    client = APIClient()
+    client.force_authenticate(user)
+    return client
+
+
+def create_customer():
```
bogdanpetrea: This is not marked with `@pytest.fixture()`. All these fixtures return more than one object, although the name suggests a single object. Also I think you can remove `create_` from their names.
You should also try to be consistent, some of these fixtures return, some don't.

stats/stats.py
```diff
@@ -0,0 +1,363 @@
...
+            self._append_element_to_key(tuple_key, granulated_stats, stats_details)
+
+    # returns the formatted granulation date according to the selected granulation parameter
+    def _get_time_granulations(self, document, granulation_field):
+        if 'time_granulation_interval' in self.granulations \
+                and 'time_granulation_interval' is not None:
+            if self.granulations['time_granulation_interval'] == 'year':
+                return getattr(document, granulation_field).strftime('%Y')
+            elif self.granulations['time_granulation_interval'] == 'month':
+                return getattr(document, granulation_field).strftime('%b %Y')
```
bogdanpetrea: We need to return timestamps here.

stats/stats.py
```diff
@@ -0,0 +1,363 @@
...
+
+            if 'granulation_field' in self.granulations:
+                granulation_value = billing_log.subscription.plan.name
+            else:
+                granulation_value = None
+
+            stats_details = {
+                "id": billing_log.subscription.id,
+                "total": billing_log.total,
+                "billing_date": billing_log.billing_date.strftime('%m/%d/%Y')
```
bogdanpetrea: I think we can use `isoformat` here.

bogdanpetrea:
Closing this as it is very unlikely it will get merged.

PR #546, opened 2017-08-22
Fix link header

vtemian:


setup.py
```diff
@@ -62,6 +62,9 @@ def read(fname):
     except IOError:
         return ''
 
+print "#" * 20
```
AMecea: Those are intended print messages?

vtemian:
Small commit from previous debuging sessions

PR #553, opened 2017-09-24
Silver cycles refactoring.

bogdanpetrea:
Introduced a couple of fields to `BillingLog`:
- `plan_billed_up_to` - The date up to which the plan has been billed.
- `metered_featuress_billed_up_to` - The date up to which the metered features have been billed.

A migration has been written to fill this fields for existing `BillingLogs`, but expect those values not to be exact when it comes to billing cycles involving trial periods. It's recommended to double check when generating the next wave of billing documents.

Two fields have been added to the `Plan` model:
- `generate_documents_on_trial_end`- If this is set to True, then billing documents will be generated when the subscription trial ends, instead of waiting for the end of the billing cycle.
- `separate_cycles_during_trial`- If this is set to True, then the trial period cycle will be split if it spans across multiple billing intervals.

The billing documents generation behaviour has changed slightly according to the aforementioned fields and it should now work for all types of intervals (used to work only for 1 month billing cycles) but it has not been thoroughly tested.


silver/documents_generator.py
```diff
@@ -89,145 +91,172 @@ def _log_subscription_billing(self, document, subscription):
...
         # Select all the active or canceled subscriptions
         subs_to_bill = []
         criteria = {'state__in': [Subscription.STATES.ACTIVE,
                                   Subscription.STATES.CANCELED]}
         for subscription in customer.subscriptions.filter(**criteria):
             if subscription.should_be_billed(billing_date):
                 subs_to_bill.append(subscription)
-                if (force_generate and subscription.state == Subscription.STATES.ACTIVE):
+
+                if force_generate and subscription.state == Subscription.STATES.ACTIVE:
```
vtemian: Why?

silver/documents_generator.py
```diff
@@ -89,145 +91,172 @@ def _log_subscription_billing(self, document, subscription):
...
                     subscription.cancel(when=Subscription.CANCEL_OPTIONS.NOW)
                     subscription.save()
+
             elif force_generate and subscription.state == Subscription.STATES.CANCELED:
                 subs_to_bill.append(subscription)
+
             elif force_generate and subscription.state == Subscription.STATES.ACTIVE:
```
vtemian: Duplicate from above.

silver/documents_generator.py
```diff
@@ -89,145 +91,172 @@ def _log_subscription_billing(self, document, subscription):
...
+
+            # This cycle decision, based on cancel_date, should be moved into `cycle_start_date` and
+            # `cycle_end_date`
+            if subscription.cancel_date:
+                relative_end_date = min(subscription.cancel_date, relative_end_date)
+
+            # Bill the plan amount
+            if plan_billed_up_to < relative_start_date:
+                if subscription.on_trial(relative_start_date):
+                    total += subscription._add_plan_trial(start_date=relative_start_date,
```
vtemian: Check what will happen when `separate_cycle_billing_trial` is `True`.

silver/documents_generator.py
```diff
@@ -89,145 +92,172 @@ def _log_subscription_billing(self, document, subscription):
...
 
-    def _create_document(self, provider, customer, subscription, billing_date):
+    def add_subscription_cycles_to_document(self, billing_date, metered_features_billed_up_to,
+                                            plan_billed_up_to, subscription,
+                                            proforma=None, invoice=None):
+        relative_start_date = metered_features_billed_up_to + ONE_DAY
+        plan_now_billed_up_to = plan_billed_up_to
+        metered_features_now_billed_up_to = metered_features_billed_up_to
+
+        prebill_plan_amount = subscription.plan.prebill_plan_amount
```
vtemian: Maybe `should_prebill_plan_amount`

silver/documents_generator.py
```diff
@@ -89,145 +92,172 @@ def _log_subscription_billing(self, document, subscription):
...
+                    relative_start_date <= subscription.trial_end <= relative_end_date):
+                relative_end_date = subscription.trial_end
+
+            # This cycle decision, based on cancel_date, should be moved into `cycle_start_date` and
+            # `cycle_end_date`
+            if subscription.cancel_date:
+                relative_end_date = min(subscription.cancel_date, relative_end_date)
+
+            # Bill the plan amount
+            if ((prebill_plan_amount and plan_billed_up_to < relative_start_date) or
```
vtemian: A comment explaining why do we need this condition would be useful.

silver/models/plans.py
```diff
@@ -63,7 +63,21 @@ class INTERVALS(object):
                   'customer to this plan.',
         verbose_name='Trial days'
     )
-
+    generate_documents_on_trial_end = models.BooleanField(
```
vtemian: Let's have a field in provider as well.

silver/models/subscriptions.py
```diff
@@ -91,23 +92,22 @@ def clean(self):
...
             if not self.start_date:
                 self.start_date = self.subscription.bucket_start_date()
             if not self.end_date:
                 self.end_date = self.subscription.bucket_end_date()
-            super(MeteredFeatureUnitsLog, self).save(force_insert, force_update,
-                                                     using, update_fields)
+            super(MeteredFeatureUnitsLog, self).save(*args, **kwargs)
 
-        if self.id:
+        else:
```
vtemian: Let's remove this else block.

PR #572, opened 2017-11-08
Add totals to documents

bogdanpetrea:


silver/models/documents/document.py
```diff
@@ -28,16 +28,23 @@ def __init__(self, *args, **kwargs):
 
     @property
     def total(self):
+        if self._total:
```
bogdanpetrea: `if self._total is not None`

silver/migrations/0042_compute_totals_in_document_view.py
```diff
@@ -0,0 +1,103 @@
...
+            model_name='proforma',
+            name='_total',
+            field=models.DecimalField(blank=True, decimal_places=2, max_digits=19, null=True),
+        ),
+        migrations.AddField(
+            model_name='proforma',
+            name='_total_in_transaction_currency',
+            field=models.DecimalField(blank=True, decimal_places=2, max_digits=19, null=True),
+        ),
+        migrations.AddField(
```
bogdanpetrea: Remove these `Document` fields.

PR #573, opened 2017-11-13
Move Invoice and Proforma into a single DB table.

bogdanpetrea:


silver/models/documents/base.py
```diff
@@ -69,6 +69,10 @@ def documents_pdf_path(document, filename):
 
 
 class BillingDocumentQuerySet(models.QuerySet):
+    def create(self, **kwargs):
```
vtemian: I guess that there is no need for this one.

PR #574, opened 2017-11-14
Don't access directly django storage's url

vtemian:


silver/models/documents/entries.py
```diff
@@ -74,17 +74,17 @@ def total_in_transaction_currency(self):
 
     @property
     def total_before_tax_in_transaction_currency(self):
-        result = self.total_before_tax * self.document.transaction_xe_rate
+        result = self.total_before_tax * (self.document.transaction_xe_rate or 1)
```
bogdanpetrea: These should not be inside this PR.

silver/api/views/documents_views.py
```diff
@@ -361,3 +363,13 @@ def get_queryset(self):
             .prefetch_related('proforma_transactions__payment_method')
 
         return (invoices | proformas).select_related('customer', 'provider', 'pdf')
+
+
+class PDFRetrieve(generics.RetrieveAPIView):
+    permission_classes = (permissions.IsAuthenticated,)
```
bogdanpetrea: I think IsStaff would be a better permission in this case.

silver/api/serializers/common.py
```diff
@@ -49,4 +50,8 @@ def create(self, validated_data):
 
 class PDFUrl(serializers.HyperlinkedRelatedField):
     def get_url(self, obj, view_name, request, format):
-        return request.build_absolute_uri(obj.pdf.url) if (obj.pdf and obj.pdf.url) else None
+        if obj.pdf is None:
+            return None
+
+        return self.reverse(view_name, kwargs={'pdf_pk': obj.pdf.pk},
```
bogdanpetrea: We should have a setting to decide between the old behaviour and this one.

silver/tests/spec/test_invoice.py
```diff
@@ -85,7 +85,7 @@ def test_post_invoice_without_invoice_entries(self):
             "state": invoice.state,
             "proforma": None,
             "invoice_entries": [],
-            "pdf_url": None,
+            "pdf_url": "http://testserver/pdfs/%s/" % invoice.pdf.pk if invoice.pdf else None,
```
bogdanpetrea: I think it would be better to know certainly what the `pdf_url` will be.

silver/tests/spec/test_invoice.py
```diff
@@ -195,7 +195,7 @@ def test_get_invoice(self):
                 "state": "issued",
                 "proforma": "http://testserver/proformas/%s/" % invoice.related_document.pk,
                 "invoice_entries": [],
-                "pdf_url": invoice.pdf.url,
+                "pdf_url": None,
```
bogdanpetrea: A test with the url of the silver pdf-view would be nice, if not one with a direct url to the pdf file too.

PR #577, opened 2017-11-15
Fix PDF on Django 1.11

vtemian:


requirements/common.txt
```diff
@@ -15,13 +15,14 @@ django-model-utils
 django-annoying
 pytz
 PyJWT
+
 # python-dev is required for xhtml2pdf to work
-xhtml2pdf
-django-xhtml2pdf
+git+https://github.com/xhtml2pdf/xhtml2pdf.git@3cb029f2cc#egg=xhtml2
```
bogdanpetrea: How about one of these https://github.com/xhtml2pdf/xhtml2pdf/releases?

Dockerfile
```diff
@@ -26,6 +26,7 @@ RUN set -ex \
         zlib \
         ca-certificates wget \
         openssl \
+        git \
```
bogdanpetrea: This should go...

PR #580, opened 2017-11-20
Improved tests when running them inside a different application.

bogdanpetrea:


silver/models/payment_methods.py
```diff
@@ -41,6 +41,7 @@ class PaymentMethod(models.Model):
     class PaymentProcessors:
         @classmethod
         def as_choices(cls):
+            print("PAYMENT_PROCESSORS", settings.PAYMENT_PROCESSORS)
```
vtemian: Debug print.

silver/models/payment_methods.py
```diff
@@ -167,9 +168,9 @@ def clean_with_previous_instance(self, previous_instance):
 
     def full_clean(self, *args, **kwargs):
         previous_instance = kwargs.pop('previous_instance', None)
+        print list(self._meta.fields[1].choices)
```
vtemian: Debug print.

PR #597, opened 2018-03-23
Add meta description to silver docs

vasiioandre:


bogdanpetrea:
It would be nice if you could squash these commits @vasiioandre .

Issue #529, opened 2017-03-23
Postgres Migration bug


macmichael01:
When running migrations 32 and 34, there is custom RunSQL command with containing AS `kind`
This for some reason breaks when running ./manage.py migrate 

The fix is to simply remove the quotation marks around kind.

sww314:
Migrations are broken for Postgres. 
These migrations all have the issue with 'kind' in quotes. 

- silver/migrations/0032_auto_20170201_1342.py
- silver/migrations/0034_auto_20170203_1644.py
- silver/migrations/0036_auto_20170514_1627.py

Not sure if it related, but in: `silver.0037_auto_20170719_1159`
`django.db.utils.NotSupportedError: cannot alter type of a column used by a view or rule
DETAIL:  rule _RETURN on view silver_document depends on column "archived_customer"`

bogdanpetrea:
Should be closed in https://github.com/silverapp/silver/commit/61ee6ea3f2a397afeb1b48a8254a3a7022405bdb.

PR #605, opened 2018-05-25
Make migrations run on PostGres

sww314:
Removes the quotes around kind in several places.
Drop and recreate view after altering fields in the view.


sww314:
Fixes #529

PR #619, opened 2018-08-16
Python2 and Python3 compatibility.

macmichael01:
Since future versions of Django will have a python3 requirement, I would attempt to make Silver Python2 and Python3 compatible. Please review!


silver/tests/commands/test_generate_docs.py
```diff
@@ -1305,7 +1307,7 @@ def test_canceled_subscription_with_trial_and_consumed_metered_features_draft(se
                                   generate_after=120, enabled=True,
                                   trial_period_days=7, amount=Decimal('200.00'),
                                   metered_features=[metered_feature])
-        start_date = dt.date(2015, 02, 01)
+        start_date = dt.date(2015, 0o2, 0o1)
```
bogdanpetrea: These weren't meant to be octals, sorry for that. The extra `0` was simply used for padding.

bogdanpetrea:
Hi @macmichael01,

This looks like the right approach to me.

Please rebase from [master](https://github.com/silverapp/silver/tree/master) so that linting and testing may run. Although I left some comments, it will be easier for me to review your PR then.

Thanks!

bogdanpetrea:
Forgot to mention you can remove the `allow_failure` for python3.6 from [the travis build](https://github.com/silverapp/silver/blob/master/.travis.yml#L16).

For [the drone build](https://github.com/silverapp/silver/blob/master/.drone.yml) it should be something similar to https://github.com/drone/drone-python/blob/master/.drone.yml, but it might require some tinkering. You can leave that to me if you want.

silver/tests/api/test_invoice.py
```diff
@@ -496,8 +503,8 @@ def test_issue_invoice_with_default_dates(self):
             'state': 'issued'
         }
         self.assertEqual(response.status_code, status.HTTP_200_OK)
-        self.assertTrue(all(item in response.data.items()
-                        for item in mandatory_content.iteritems()))
+        self.assertTrue(all(item in list(response.data.items())
+                        for item in six.iteritems(mandatory_content)))
```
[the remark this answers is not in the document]
macmichael01: I reverted back the occurrences of six.iteritems since the compatibility script I used was attempting to replace model queries. Django "should" have these types of compatibilities concerns covered.

silver/models/transactions/transaction.py
```diff
@@ -295,8 +297,8 @@ def update_document_state(self):
                 not self.document.amount_to_be_charged_in_transaction_currency):
             self.document.pay()
 
-    def __unicode__(self):
-        return unicode(self.uuid)
+    def __str__(self):
```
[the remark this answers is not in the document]
macmichael01: I updated the models with the decorator. I used smart_text to wrap where unicode was previously defined. Basically tried to remove parts of the code where six was added. Were you suggesting that I remove six imports altogether? There are still some places where six is getting imported to override standard python methods.

Eg.
from six.moves import zip
from six.moves import map
from six.moves import range

silver/tests/api/test_invoice.py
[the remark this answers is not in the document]
bogdanpetrea: `dict.iteritems()` is gone from Python3, so we're either stuck with `dict.items()` which in Python2 might take up some extra memory or `six.iteritems()` which uses the same implementation of dict.items()` from Python3.

Using `dict.items()` should be fine most of the times, especially in tests and places where we iterate over relatively few objects and 'six.iteritems()` is slightly harder to read.

What I should've said in the initial comment is that we should try to be consistent if possible.

bogdanpetrea:
It seems tests are passing for Python2: https://travis-ci.org/silverapp/silver/jobs/419958428
But they're not passing for Python3: https://travis-ci.org/silverapp/silver/jobs/419958429
Linting fails in both cases, but it should be easy fixing it.

I see the failing tests are related to:
1. `exception.message` not being available anymore in Python3. Converting the exception object to string instead to obtain a message, should do the trick.
2. `'dict' object has no attribute 'iteritems'`. Use `.items()` as discussed previously.
3. `pdf_content = ContentFile(pdf_file_object)` error.
https://github.com/silverapp/silver/blob/0009ff4ca52dfc711e2f160ad90b449060fc4007/silver/models/documents/pdf.py#L41 should become `pdf_file_object = BytesIO()`.
4. `django.urls.exceptions.NoReverseMatch: Reverse for 'payment' with keyword arguments`
https://github.com/silverapp/silver/blob/3b718bb7a44c7e1e42479e9f59a0f68c841abda0/silver/utils/payments.py#L33
should become `kwargs = {'token': force_text(_get_jwt_token(transaction))}`

Let me know if you need any help.

silver/tests/unit/test_proforma.py
```diff
@@ -12,14 +12,17 @@
...
 
+from six.moves import zip
+
 from django.test import TestCase
 
 from silver.models import DocumentEntry, Invoice, Proforma
-from silver.tests.factories import ProformaFactory, DocumentEntryFactory, \
-    CustomerFactory
+from silver.tests.factories import (
+    ProformaFactory, DocumentEntryFactory, CustomerFactory)
```
bogdanpetrea: Style: Please put ending parenthesis ) on a newline.

silver/tests/unit/test_payments_util.py
```diff
@@ -78,16 +83,21 @@ def test_get_transaction_from_expired_token(self):
...
+        else:
+            expected_token = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ0cmFuc2FjdG' \
+                             'lvbiI6IjZmYTQ1OWVhLWVlOGEtNGNhNC04OTRlLWRiNzdlMTYwM' \
+                             'zU1ZSIsImV4cCI6MTQ5Nzk2NTY0MH0.-bpx5A3DfSe3-HO6aH_g' \
+                             'lS8adcCxUn8lSK1-RPxohhI'
         with patch('silver.utils.payments.datetime') as mocked_datetime:
-            mocked_datetime.utcnow.return_value = datetime.strptime('Jun 20 2017 1:33PM',
-                                                                    '%b %d %Y %I:%M%p')
+            mocked_datetime.utcnow.return_value = datetime.strptime(
+              'Jun 20 2017 1:33PM', '%b %d %Y %I:%M%p')
```
bogdanpetrea: Style: Please put ending parenthesis ) on a newline.

silver/tests/unit/test_payments_util.py
```diff
@@ -78,16 +83,21 @@ def test_get_transaction_from_expired_token(self):
...
-                         'lvbiI6IjZmYTQ1OWVhLWVlOGEtNGNhNC04OTRlLWRiNzdlMTYwM' \
-                         'zU1ZSIsImV4cCI6MTQ5Nzk2NTY0MH0.-bpx5A3DfSe3-HO6aH_g' \
-                         'lS8adcCxUn8lSK1-RPxohhI'
+        if six.PY3:
+            expected_token = 'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ0cmFuc2FjdG' \
+                             'lvbiI6IjZmYTQ1OWVhLWVlOGEtNGNhNC04OTRlLWRiNzdlMTYwM' \
+                             'zU1ZSIsImV4cCI6MTQ5Nzk2NTY0MH0.NuqwXlHt27nRxg5W2hHo' \
+                             'P5ugnLmw-7QYBumO7lRa1i0'
+        else:
+            expected_token = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ0cmFuc2FjdG' \
```
bogdanpetrea: I don't understand how this token remained the same even though you modified the `uuid` above. :confused:

silver/tests/api/test_invoice.py
```diff
@@ -166,8 +171,10 @@ def test_get_invoice(self, mocked_settings):
                                                         [transaction.customer.id])),
             "provider": build_absolute_test_url(reverse('provider-detail',
                                                         [transaction.provider.id])),
-            "amount": "%s.00" % str(transaction.amount),
-            "currency": "USD",
+            "amount": (
```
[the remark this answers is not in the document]
macmichael01: oh ha. that was silly of me.

silver/tests/commands/test_generate_docs.py
```diff
@@ -1426,7 +1428,7 @@ def test_canceled_subscription_with_trial_and_trial_underflow(self):
...
```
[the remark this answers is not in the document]
macmichael01: I didn't catch that one. Oddly the unit tests pass. I guess there may need more coverage for this one.

silver/tests/unit/test_payments_util.py
macmichael01: Yeah I didn't quite understand that either. I just figured it was a typo.

silver/admin.py
```diff
@@ -60,15 +65,15 @@ def metadata(obj):
...
+        for key, value in obj.meta.items():
             d += u'%s: <code>%s</code><br>' % (escape(key), escape(value))
     return d
 metadata.allow_tags = True
 
 
 def tax(obj):
     return ("{} {:.2f}%".format(obj.sales_tax_name, obj.sales_tax_percent)
-            if obj.sales_tax_percent > 0 else '')
+            if obj.sales_tax_percent is not None and obj.sales_tax_percent > 0 else '')
```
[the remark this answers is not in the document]
bogdanpetrea: Well if `obj.sales_tax_percent` is `True`, it means that `obj.sales_tax_percent is not None` and also that `obj.sales_tax_percent != 0`. As `sales_tax_percent` is already known to be `>= 0`, it means that `obj.sales_tax_percent` is equivalent to `obj.sales_tax_percent is not None and obj.sales_tax_percent > 0` in this case.

silver/tests/unit/test_payments_util.py
bogdanpetrea: The collision is probably due to the weak secret that is being used
```python
@override_settings(PAYMENT_METHOD_SECRET='a')
```

bogdanpetrea: `timedelta(minutes=1)` is equivalent to `timedelta(seconds=60)`. It's probably the just the weak secret value. We can address that some other time...

silver/models/documents/base.py
```diff
@@ -34,32 +35,31 @@
...
-
 PAYMENT_DUE_DAYS = getattr(settings, 'SILVER_DEFAULT_DUE_DAYS', 5)
 
 logger = logging.getLogger(__name__)
 
 
 def documents_pdf_path(document, filename):
     path = '{prefix}{company}/{doc_name}/{date}/{filename}'.format(
-        company=slugify(unicode(
+        company=slugify(force_text(
```
bogdanpetrea: `force_text` is not imported in this file and will cause an error when this method is called.

silver/tests/commands/test_generate_docs.py
```diff
@@ -1445,7 +1447,7 @@ def test_canceled_subscription_with_trial_and_trial_underflow(self):
         )
 
         subscription.cancel(when=Subscription.CANCEL_OPTIONS.END_OF_BILLING_CYCLE)
-        subscription.cancel_date = dt.date(2015, 02, 28)
+        subscription.cancel_date = dt.date(2015, 0o2, 28)
```
bogdanpetrea: These should not be octals, as in `0o2` should simply be `2`. I will change them if you can't be bothered.

macmichael01: Agh. That script I ran did this. I ran a quick search and found a couple more dates with 0o. It doesn't appear to cause any harm when creating the actual dates. Should be fixed now.

silver/models/documents/base.py
macmichael01: Good catch.

PR #644, opened 2019-05-29
Remove the unneccessary old requirement (xhtml2pdf in dependency_links)

shurph:
Currently a fresh version of `xhtml2pdf` installs from the `requirements/common.txt` file.

So, it seems like xhtml2pdf in dependency_links is some redundant part.


bogdanpetrea:
Good catch!

Issue #645, opened 2019-06-06
Initial Migration Failing Django 2.1.3 / Postgres 11.2


ajaff:
Hi,

I am new to using Silver - I just downloaded the silver app and tried to run the migration however I am receiving the below error.

**Django 2.1.3
PostgreSQL 11.2**

```

Operations to perform:
  Apply all migrations: silver
Running migrations:
  Applying silver.0001_initial... OK
  Applying silver.0002_auto_20150416_1009... OK
  Applying silver.0003_auto_20150417_0634... OK
  Applying silver.0004_auto_20150429_1731... OK
  Applying silver.0005_auto_20150429_1732... OK
  Applying silver.0006_auto_20150430_1457... OK
  Applying silver.0007_auto_20150430_1549... OK
  Applying silver.0008_auto_20150430_1804... OK
  Applying silver.0009_auto_20150717_1428... OK
  Applying silver.0031_auto_20170125_1343...Traceback (most recent call last):
  File "C:\Users\Ahmed\OneDrive\Documents\RESX\Code\tornado\venv\lib\site-packages\django\db\backends\utils.py", line 85, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.StringDataRightTruncation: value too long for type character varying(4)


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "manage.py", line 16, in <module>
    execute_from_command_line(sys.argv)
  File "C:\Users\Ahmed\OneDrive\Documents\RESX\Code\tornado\venv\lib\site-packages\django\core\management\__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "C:\Users\Ahmed\OneDrive\Documents\RESX\Code\tornado\venv\lib\site-packages\django\core\management\__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "C:\Users\Ahmed\OneDrive\Documents\RESX\Code\tornado\venv\lib\site-packages\django\core\management\base.py", line 316, in run_from_argv
    self.execute(*args, **cmd_options)
  File "C:\Users\Ahmed\OneDrive\Documents\RESX\Code\tornado\venv\lib\site-packages\django\core\management\base.py", line 353, in execute
    output = self.handle(*args, **options)
  File "C:\Users\Ahmed\OneDrive\Documents\RESX\Code\tornado\venv\lib\site-packages\django\core\management\base.py", line 83, in wrapped
[...]

PR #646, opened 2019-06-07
Django 2.2 support

shurph:
These changes provide  an experimental Django 2.2 support.

And there are few fixes for #645 issue.


shurph:
@bogdanpetrea , could you please look at this when you  will have a chance?

silver/tests/unit/test_pdf.py
```diff
@@ -86,14 +86,14 @@ def test_pdf_mark_as_clean_mimic_2_threads():
 
 
 @pytest.mark.django_db
-def test_pdf_mark_as_dirty_min_value():
-    # somehow inconsistent state
-    pdf = PDF.objects.create(dirty=-1)
+def test_pdf_mark_as_clean_min_value():
```
bogdanpetrea: Why did this test have to be changed?

bogdanpetrea:
Also, you should add Django 2.2 to the CI builds:
https://github.com/silverapp/silver/blob/master/.drone.yml#L55
https://github.com/silverapp/silver/blob/master/.travis.yml#L11

silver/tests/unit/test_pdf.py
shurph: The main purpose to fit to the new feature in Django 2.2 — extended constraints.
With 2.2 you can't apply `pdf = PDF.objects.create(dirty=-1)` because integrity error will be produced (`dirty` is PositiveIntegerField).

As I understand `test_pdf_mark_as_dirty_min_value` was written to be sure that `dirty` can not be below 0. So it seems like with the new 2.2 feature and also with you changes (https://github.com/silverapp/silver/commit/e4dd96ef44bf62f52a7f939a213af4e6bad1ed6a) the test_pdf_mark_as_clean_min_value is unnecessary, test_pdf_mark_as_clean_min_value is enough.

shurph:
@bogdanpetrea I've added Django 2.2 to CI builds and also switched travis default image (Ubuntu 14.04) to the Ubuntu 16.04 because Django 2.2 requires a fresh SQLite version.

silver/tests/unit/test_pdf.py
bogdanpetrea: Makes sense. Thanks!

shurph:
@bogdanpetrea thank you for merging this PR. Couls you please release `0.8b2` with these changes?

PR #655, opened 2019-07-19
Split docs pages

vasiioandre:
This is necessary to make the Algolia search work.


bogdanpetrea:
I agree we need to make the docs search work, but not by splitting the documentation into arbitrary parts.

PR #668, opened 2019-12-27
Fixed bugs in admin views

mbhassan:
Fixed bugs in *autocomplete views  where self.request.user.is_authenticated() was being called instead of being accessed as an attribute.


bogdanpetrea:
Thanks @mbhassan!

Issue #676, opened 2021-01-09
Considering contributring


adamcharnock:
Hi @bogdanpetrea,

I've come across silver when looking for a way to integration multiple payment providers into Django. I'm very my interested in contriburing, but I wanted to check-in first to say hi 👋 

My motivation is that we're am starting an ISP in Portugal, and a need a billing system which can work with European payment methods and be extendable in future. We are going to start with Stripe, but we don't want to be locked into them. Payment methods we'll be interested in are card, SEPA debit, and MultiBanco.

I have the silver codebase setup locally and all the tests pass. I've listed below a few tasks that would definitely interest me, but could you give me a very rough idea of how you would feel about receiving PRs in each of these cases?

* Creation of Stripe payment processor (as a separate project)
* Adding Django 3 support
* Adding a `pyproject.toml` so poetry can be used
* Adding type hints (and therefore dropping Python 3.5 support)
* Applying black code formatting
* Probably additional (as yet to be figured out) settings to tweak silver's behaviour, while *not* modifying the default behaviour.

Also, are you still using Silver within Presslabs? Will there be resources to review PRs and perform releases for the foreseeable?

Sorry for all the questions, I just thought it best to check up-front. I'm definitely keen to contribute!

Also, I've been working with Django and Python for a long time now. Take a look at my profile if you want to size me up, also [my website](https://adamcharnock.com/) as well as [the Lightbus project](https://lightbus.org) as an example of an OSS project of mine.

adamcharnock:
I ended up making a start on this regardless, it seems I couldn't resist 😄 

I may take a crack at the Stripe integration today.

bogdanpetrea:
Hi @adamcharnock , I was out of office but starting today I will be able to assist you.

All the points you've mentioned sound good to me! And yes we're still making good use of Silver at Presslabs.

I appreciate your enthusiasm and will soon review the PRs you've sent.

adamcharnock:
Hi @bogdanpetrea,

Thank you for getting back to me, no replies expected on the weekend!

As I started to try to integrate with Silver over the weekend I started to realise that it probably more sense for me to develop something simpler in-house. Our recurring invoicing is handled by another internal system, so all we really need is a way to setup payment methods for future use, and then a way to take payments. Given that, I think I have an in-house system mostly working now.

So I'm sorry to say won't be getting more involved with Silver, but I definitely would if we needed more of its functionality.

Regardless, I hope the Django 3 PR (#677) is still useful.

[...]

PR #677, opened 2021-01-09
Adding Django 3 support

adamcharnock:
As per #676, I've gone ahead and brought all dependencies up to date and got all tests passing. Note:

* Running against Django 3.1.5
* Fixed various warnings
* Now using Django's internal JSONField
* Removed all Python 2 compatibility imports, which are largely gone from Django 3 anyway
* I managed my own environment using poetry, but I haven't committed those files. I've updated the various requirements files with the new versions, so CI *should* pass.

I'm sorry that this this is a big PR, but I don't really think there a practical way to do this incrementally. I've tried to make every commit pretty stand-alone though, so you should be able to walk through the individual commits and see what is going on. It's mostly very simple changes over a large number of files.


adamcharnock:
It looks like the generic (not postgres-specific) `JSONField` is only in Django 3. So it looks like keeping Django <3 support is going to require keeping a third-party JSON field library in the requirements.

So, forgive me for asking, but is there any appetite for doing a major (backwards-compatible) release which drops support Django < 3, and older Python versions too?

bogdanpetrea:
> It looks like the generic (not postgres-specific) `JSONField` is only in Django 3. So it looks like keeping Django <3 support is going to require keeping a third-party JSON field library in the requirements.
> 
> So, forgive me for asking, but is there any appetite for doing a major (backwards-compatible) release which drops support Django < 3, and older Python versions too?

No, Django > 3 and Python >= 3.6 should be the way to go forward. I'm only concerned about migrating existing data from the third-party JSONField to the Django in-built one.

bogdanpetrea:
Hey @adamcharnock , I've changed our Drone config and the Travis one to use Django>3 and Python>3.6. Please rebase your PR branch onto master.

adamcharnock:
@bogdanpetrea Rebase done

bogdanpetrea:
Included and merged in a separate PR (https://github.com/silverapp/silver/pull/680). Thanks again!

